PR #5, opened 2020-09-17
Add README and documentation

viswajithiii:
* Add README and documentation.
* Write commands to print out the list of checks and templates, both plain and in markdown format. The markdown docs are checked in to the repo for better documentation.


README.md
```diff
@@ -1 +1,34 @@
...
+`kube-linter` binaries can be downloaded from [the Releases page](https://github.com/stackrox/kube-linter/releases).
+Download the `kube-linter` binary, and add it to your PATH.
+
+## Usage
+
+To lint directories or files, simply run `./kube-linter lint files_or_dirs ...`. If a directory is passed, all files
+with `.yaml` or `.yml` extensions are parsed, and Kubernetes objects are loaded from them. If a file is passed,
+it is parsed irrespective of extension.
+
+Users can pass a config file using the `--config` file to control which checks are executed, and to configure custom checks.
```
misberner, line 26: `--config` *flag*

docs/README.md
```diff
@@ -0,0 +1,38 @@
...
+You can run `kube-linter --help` to see a list of supported commands and flags. For each subcommand, you can
+run `kube-linter <subcommand> --help` to see detailed help text and flags for it.
+
+## Running the linter
+
+To lint directories or files, simply run `./kube-linter lint files_or_dirs ...`. If a directory is passed, all files
+with `.yaml` or `.yml` extensions are parsed, and Kubernetes objects are loaded from them. If a file is passed,
+it is parsed irrespective of extension.
+
+Users can pass a config file using the `--config` file to control which checks are executed, and to configure custom checks.
```
misberner, line 17: flag again (is there some way to not duplicate the contents between this and the top-level readme)?

README.md
```diff
@@ -1 +1,34 @@
...
+
+The list of checks that is run is configurable. `kube-linter` comes with several built-in checks, only some of which
+are enabled by default. Users can also create custom checks.
+
+## Install
+
+If you have `go` installed, you can run `go get golang.stackrox.io/kube-linter/cmd/kube-linter`.
+
+`kube-linter` binaries can be downloaded from [the Releases page](https://github.com/stackrox/kube-linter/releases).
+Download the `kube-linter` binary, and add it to your PATH.
```
misberner, line 18: IMHO a readme in an open source repository should always include custom build instructions and contributing guidelines (or references to these)

internal/command/checks/command.go
```diff
@@ -0,0 +1,90 @@
...
+	dashes = func() string {
+		var sb strings.Builder
+		for i := 0; i < 30; i++ {
+			sb.WriteRune('-')
+		}
+		return sb.String()
+	}()
+
+	formatsToRenderFuncs = map[string]func([]check.Check, io.Writer){
+		common.PlainFormat: func(checks []check.Check, out io.Writer) {
```
misberner, line 28: Move these to separate functions instead of inlining them. I would also encourage you to use Go templates for formatting. Especially the markdown formatting (a) looks ugly in code and (b) is extremely limited as you cannot add any custom styling, notes, etc.
(You might need to add either a custom function or translate the checks into a separate type to realize the "enabled by default" flag)

internal/command/checks/command.go
```diff
@@ -0,0 +1,90 @@
...
+var (
+	dashes = func() string {
+		var sb strings.Builder
+		for i := 0; i < 30; i++ {
+			sb.WriteRune('-')
+		}
+		return sb.String()
+	}()
+
+	formatsToRenderFuncs = map[string]func([]check.Check, io.Writer){
```
misberner, line 27: Possibly irrelevant (see below), but the brace following the closing parens without space in between looks weird - doesn't the linter flag this?

internal/command/common/format_wrapper.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// This block enumerates all supported formats.
+var (
+	PlainFormat    = newFormat("plain")
+	MarkdownFormat = newFormat("markdown")
+)
+var (
+	// AllSupportedFormats contains the list of all supported formats.
+	AllSupportedFormats = set.NewStringSet()
```
misberner, line 17: Does Go have some form of dependency resolution when calling global variable initializers? To me it looks like there is nothing ensuring that this initialization has taken place at them time when `newFormat("plain")` above calls `AllSupportedFormats.Add`.

It also seems weird that we store a set of supported formats. That should be a derived property, derived from a map mapping the IDs to something semantic like a render function. In short - why not derive this from `formatToRenderFunc`?

internal/command/templates/command.go
```diff
@@ -0,0 +1,90 @@
...
+var (
+	dashes = func() string {
+		var sb strings.Builder
+		for i := 0; i < 30; i++ {
+			sb.WriteRune('-')
+		}
+		return sb.String()
+	}()
+
+	formatsToRenderFuncs = map[string]func([]check.Template, io.Writer){
```
misberner, line 26: See above

misberner (changes requested):

README.md
connorgorman: 👍

docs/README.md
viswajithiii: Fixed. And I just removed the duplicate content from the top-level README in favour of just linking.

internal/command/common/format_wrapper.go
viswajithiii: > Does Go have some form of dependency resolution when calling global variable initializers? To me it looks like there is nothing ensuring that this initialization has taken place at them time when newFormat("plain") above calls AllSupportedFormats.Add.

Yup, Go handles this really well. See https://golang.org/ref/spec#Program_initialization_and_execution.

> In short - why not derive this from formatToRenderFunc?

The problem is that it's shared between the `checks` and `templates` command, which have their own `formatToRenderFunc`s. I've written unit tests to ensure the keys of that map and this are in sync.

README.md
viswajithiii: Added instructions for building from source. Contributing guidelines, will do in a follow-up -- we still need to iron those out.

internal/command/checks/command.go
viswajithiii: That brace is actually consistent with `go fmt` since it applies to the `map`, and not the `func` (although I agree it pattern matches with braces around a `func` and looks weird).

internal/command/checks/command.go
viswajithiii: Done. Also moved to templates.

internal/command/checks/command.go
```diff
@@ -0,0 +1,106 @@
...
+		return sb.String()
+	}()
+
+	formatsToRenderFuncs = map[string]func([]check.Check, io.Writer) error{
+		common.PlainFormat:    renderPlain,
+		common.MarkdownFormat: renderMarkdown,
+	}
+)
+
+func renderPlain(checks []check.Check, out io.Writer) error { //nolint:unparam // The function signature is required to match formatToRenderFuncs
```
misberner, line 33: Why don't you use a Go template for this?

internal/command/checks/command.go
```diff
@@ -0,0 +1,106 @@
...
+			chk.Name, chk.Description, chk.Template, chk.Params, defaultchecks.List.Contains(chk.Name))
+		if i != len(checks)-1 {
+			fmt.Fprintf(out, "\n%s\n\n", dashes)
+		}
+	}
+	return nil
+}
+
+const (
+	markDownTemplateStr = `The following table enumerates built-in checks:
```
misberner, line 45: Ok so what I actually had in mind was something like `--format go-template="{{ template str }}"` or `--format go-template=@file`. It is fairly established in the Go world to allow users to specify Go templates for all "print"-like commands (e.g., `go list`, `kubectl get`.. etc) (although the file variant is something non-standard, but useful here).
This works for the initial iteration but having to resort to `{{backtick}}` is.. ugly..

internal/command/common/markdown.go
```diff
@@ -0,0 +1,24 @@
...
+
+// MustInstantiateTemplate instanties the given go template with a common list of
+// functions. It panics if there is an error.
+func MustInstantiateTemplate(templateStr string) *template.Template {
+	tpl, err := template.New("").Funcs(
+		template.FuncMap{
+			"backtick": func() string {
+				return "`"
+			},
+			"joinstrings": strings.Join,
```
misberner, line 18: tbh I'd prefer to use a standard Go template library like sprig instead, especially if you want to eventually allow users to provide their own templates

internal/command/templates/command.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"golang.stackrox.io/kube-linter/internal/check"
+	"golang.stackrox.io/kube-linter/internal/command/common"
+	"golang.stackrox.io/kube-linter/internal/templates"
+)
+
+var (
+	dashes = func() string {
```
misberner, line 17: Maybe worth adding a `stringutils.Repeat`?

internal/command/templates/command.go
```diff
@@ -0,0 +1,97 @@
...
+|
+{{ end -}}
+`
+)
+
+var (
+	markDownTemplate = common.MustInstantiateTemplate(markDownTemplateStr)
+)
+
+func renderPlain(templates []check.Template, out io.Writer) error { //nolint:unparam // The function signature is required to match formatToRenderFuncs
```
misberner, line 49: which parameter is unused? oO

viswajithiii: `error` is alwasys `nil`.

internal/command/checks/command.go
viswajithiii: This function is pretty simple, so I think it's better to keep it as is. I find normal code easier to read (plus the compile-time safety is nice).

internal/command/checks/command.go
viswajithiii: Ahh got it. Yeah, I planned to support Go template output for the `lint` command but wasn't sure if it was worth it for these commands... Will do in a follow-up.

internal/command/templates/command.go
viswajithiii: Done.

internal/command/common/markdown.go
viswajithiii: Done.

config.yaml.example
```diff
@@ -0,0 +1,19 @@
+# customChecks defines custom checks.
+customChecks:
+- name: "required-label-app"
+  template: "required-label"
+  params:
+    key: "app"
+checks:
+  # if doNotAutoAddDefaults is true, default checks are not automatically added.
+  # Otherwise, they are.
```
connorgorman, line 9: Remove, this is pretty obvious based on the previous line

connorgorman: Also, I think doNotAutoAddDefaults is backwards, should be autoAddDefaults that defaults to true

connorgorman: it's also just a bit too verbose

internal/ternary/ternary.go
```diff
@@ -0,0 +1,9 @@
+package ternary
```
connorgorman, line 1: why is this in the PR? Also, adding a ternary is a bit of a strange thing in golang as it feels like it should be a language level thing

connorgorman (approved):
Minor stuff. Looks good

internal/ternary/ternary.go
viswajithiii: Yeah I was using it initially but not anymore. Removed.

PR #6, opened 2020-09-18
Add a couple more check templates and built-in checks

viswajithiii:
Add check templates for read-only root FS and `runAsNonRoot`.

### DIFF
--- a/internal/builtinchecks/yamls/run-as-non-root.yaml
+++ b/internal/builtinchecks/yamls/run-as-non-root.yaml
@@ -1,0 +1,6 @@
+name: "run-as-non-root"
+description: "Alert on containers not set to runAsNotRoot"
+scope:
+  objectKinds:
+    - DeploymentLike
+template: "run-as-non-root"
--- a/internal/builtinchecks/yamls/run-as-non-root.yaml
+++ b/internal/builtinchecks/yamls/run-as-non-root.yaml
@@ -1,5 +1,5 @@
 name: "run-as-non-root"
-description: "Alert on containers not set to runAsNotRoot"
+description: "Alert on containers not set to runAsNonRoot"
 scope:
   objectKinds:
     - DeploymentLike
--- a/internal/defaultchecks/default_checks.go
+++ b/internal/defaultchecks/default_checks.go
@@ -9,7 +9,6 @@ var (
 	List = set.NewFrozenStringSet(
 		"privileged-container",
 		"env-var-secret",
-		"no-v1beta-deployments",
 		"no-read-only-root-fs",
 		"run-as-non-root",
 	)
--- a/internal/defaultchecks/default_checks.go
+++ b/internal/defaultchecks/default_checks.go
@@ -8,5 +8,9 @@ var (
 	// List is the list of built-in checks that are enabled by default.
 	List = set.NewFrozenStringSet(
 		"privileged-container",
+		"env-var-secret",
+		"no-v1beta-deployments",
+		"no-read-only-root-fs",
+		"run-as-non-root",
 	)
 )
--- a/internal/templates/readonlyrootfs/template.go
+++ b/internal/templates/readonlyrootfs/template.go
@@ -1,0 +1,40 @@
+package readonlyrootfs
+
+import (
+	"fmt"
+
+	"golang.stackrox.io/kube-linter/internal/check"
+	"golang.stackrox.io/kube-linter/internal/diagnostic"
+	"golang.stackrox.io/kube-linter/internal/extract"
+	"golang.stackrox.io/kube-linter/internal/lintcontext"
+	"golang.stackrox.io/kube-linter/internal/objectkinds"
+	"golang.stackrox.io/kube-linter/internal/templates"
+)
+
+func init() {
+	templates.Register(check.Template{
+		Name:        "read-only-root-fs",
+		Description: "Flag containers without read-only root file systems",
+		SupportedObjectKinds: check.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters: nil,
+		Instantiate: func(_ map[string]string) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				podSpec, found := extract.PodSpec(object.K8sObject)
+				if !found {
+					return nil
+				}
+				var results []diagnostic.Diagnostic
+				for i := range podSpec.Containers {
+					container := podSpec.Containers[i]
+					sc := container.SecurityContext
+					if sc == nil || sc.ReadOnlyRootFilesystem == nil || !*sc.ReadOnlyRootFilesystem {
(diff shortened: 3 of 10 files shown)


internal/defaultchecks/default_checks.go
```diff
@@ -8,5 +8,9 @@ var (
 	// List is the list of built-in checks that are enabled by default.
 	List = set.NewFrozenStringSet(
 		"privileged-container",
+		"env-var-secret",
+		"no-v1beta-deployments",
```
connorgorman, line 12: Does this one already exist?

internal/templates/readonlyrootfs/template.go
```diff
@@ -0,0 +1,40 @@
...
+		},
+		Parameters: nil,
+		Instantiate: func(_ map[string]string) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				podSpec, found := extract.PodSpec(object.K8sObject)
+				if !found {
+					return nil
+				}
+				var results []diagnostic.Diagnostic
+				for i := range podSpec.Containers {
```
connorgorman, line 29: Why not range with _, container?

internal/templates/runasnonroot/template.go
```diff
@@ -0,0 +1,51 @@
...
+				podSpec, found := extract.PodSpec(object.K8sObject)
+				if !found {
+					return nil
+				}
+				var results []diagnostic.Diagnostic
+				var podSCValue bool
+				if podSpec.SecurityContext != nil && podSpec.SecurityContext.RunAsNonRoot != nil {
+					podSCValue = *podSpec.SecurityContext.RunAsNonRoot
+				}
+				for i := range podSpec.Containers {
```
connorgorman, line 33: would prefer for _, container := range

connorgorman (approved):

internal/builtinchecks/yamls/run-as-non-root.yaml
```diff
@@ -0,0 +1,6 @@
+name: "run-as-non-root"
+description: "Alert on containers not set to runAsNotRoot"
```
misberner, line 2: should this be `runAsNo*n*Root`?

internal/templates/readonlyrootfs/template.go
misberner: IIRC linter complaining about large object for loop iteration variable. Though since you do `container := podSpec.Containers[i]`, really nothing is gained here. So either change the assignment to `:= &podSpec.Containers[i]`, or add a nolint directive and apply gorman's suggestion

internal/templates/readonlyrootfs/template.go
```diff
@@ -0,0 +1,40 @@
...
+					container := podSpec.Containers[i]
+					sc := container.SecurityContext
+					if sc == nil || sc.ReadOnlyRootFilesystem == nil || !*sc.ReadOnlyRootFilesystem {
```
misberner, line 32: no action required: it would really be nice if we could specify checks in a simpler language where, e.g., we would have an `?.` operator

internal/templates/runasnonroot/template.go
```diff
@@ -0,0 +1,51 @@
...
+		Parameters: nil,
+		Instantiate: func(_ map[string]string) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				podSpec, found := extract.PodSpec(object.K8sObject)
+				if !found {
+					return nil
+				}
+				var results []diagnostic.Diagnostic
+				var podSCValue bool
+				if podSpec.SecurityContext != nil && podSpec.SecurityContext.RunAsNonRoot != nil {
```
misberner, line 30: I wonder if it wouldn't be simpler to work with `unstructured` objects in order to allow nested and uniform (between pod/container) element access...

internal/templates/runasnonroot/template.go
```diff
@@ -0,0 +1,51 @@
...
+					if container.SecurityContext != nil {
+						runAsNonRoot = container.SecurityContext.RunAsNonRoot
+					}
+					if runAsNonRoot != nil && *runAsNonRoot {
+						continue
+					}
+					if runAsNonRoot == nil && podSCValue {
+						continue
+					}
+					results = append(results, diagnostic.Diagnostic{Message: fmt.Sprintf("container %q is not set to runAsNonRoot", container.Name)})
```
misberner, line 45: This check seems overzealous as it flags deployments where (a) the container image uses a non-root `USER` and (b) a non-0 UID is specified. You could argue that it's good practice to enforce non-root via the SC nonetheless, but eh, IMHO there's more harm than good being done by insisting on principles to such an extent. If you really want to ship it as enabled by default, you should check for at least (b).

misberner (changes requested):

internal/defaultchecks/default_checks.go
viswajithiii: Ah, good catch. That's on another branch now, initially had it here, but decided to refactor it before creating a PR. Removed it from this list and added a unit test to make sure all these check names are valid.

internal/builtinchecks/yamls/run-as-non-root.yaml
viswajithiii: Fixed.

internal/templates/runasnonroot/template.go
viswajithiii: Yeah, I've been thinking about that too. Still deciding whether to refactor it. The type-safety is definitely nice, but it doesn't feel _super_ helpful given that you need to be so careful with the `nil` dereferences.

internal/templates/readonlyrootfs/template.go
viswajithiii: Just removed this specific lint check. Not convinced it added enough value to justify the clunkiness.

internal/templates/runasnonroot/template.go
viswajithiii: Fair enough, checked for (b).

internal/templates/runasnonroot/template.go
viswajithiii: Fixed.

internal/templates/runasnonroot/template.go
```diff
@@ -0,0 +1,72 @@
...
+		Parameters: nil,
+		Instantiate: func(_ map[string]string) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				podSpec, found := extract.PodSpec(object.K8sObject)
+				if !found {
+					return nil
+				}
+				var results []diagnostic.Diagnostic
+				for _, container := range podSpec.Containers {
+					runAsNonRoot := effectiveRunAsNonRoot(podSpec.SecurityContext, container.SecurityContext)
```
misberner, line 51: This should be evaluated only after l.56

misberner (approved):
Minor suggestion to move a couple of lines around, LG o/w

PR #7, opened 2020-09-21
Build binaries for all platforms; add version command

viswajithiii:


get-tag
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+set -e
```
misberner, line 3: This has no effect, remove

get-tag
```diff
@@ -0,0 +1,4 @@
...
+git describe --tags --abbrev=10 --dirty
```
misberner, line 4: Note that this might suffer from the issue that if you re-run the last master build after tagging it, it will now be run with the _tag_ as the version. You'd need to pass `--long` here and check for `CIRCLE_TAG` to avoid this issue.

scripts/variable_stamps.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+source "${SCRIPT_DIR}/utils.sh"
+
+set -euo pipefail
+
+gitroot="$(git rev-parse --show-toplevel)"
```
misberner, line 8: Isn't the location of the git root determined relative to `SCRIPT_DIR` anyway? Or are you future proofing it to be used in other repos?

misberner (changes requested):

Makefile
```diff
@@ -10,10 +10,12 @@ deps: go.mod
 	@go mod verify
 	@touch deps
 
+UNAME_S := $(shell uname -s)
+HOST_OS:=linux
```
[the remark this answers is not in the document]
viswajithiii: Done.

Makefile
```diff
@@ -10,10 +10,12 @@ deps: go.mod
...
+ifeq ($(UNAME_S),Darwin)
+    HOST_OS:=darwin
```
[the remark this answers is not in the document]
viswajithiii: Done.

Makefile
```diff
@@ -80,7 +86,11 @@ packr: $(PACKR_BIN)
 
 .PHONY: build
 build: packr
-	go build -o ./bin/kube-linter ./cmd/kube-linter
+	@CGO_ENABLED=0 GOOS=darwin scripts/go-build.sh ./cmd/kube-linter
```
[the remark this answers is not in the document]
viswajithiii: Done.

get-tag
viswajithiii: Done.

scripts/variable_stamps.sh
viswajithiii: Yeah, it just seems like a better option since the location of `get-tag` relative to `SCRIPT_DIR` is not constrained.

get-tag
viswajithiii: DOne.

misberner (approved):

PR #8, opened 2020-09-21
Upload release artifacts when a tag is cut

viswajithiii:
Use the GitHub API to automatically create a release, and upload the binaries to it, whenever a tag is cut. The release will be marked as a pre-release; a human needs to go in and verify that all is well before promoting it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,7 @@ jobs:
         name: Create a GitHub release, if on tag.
         command: |
           if [[ -n "${CIRCLE_TAG}" ]]; then
-
+            ./scripts/create_github_release.sh "${CIRCLE_TAG}" ./bin
           fi
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,17 +57,16 @@ jobs:
     - run:
         name: Create a GitHub release, if on tag.
         command: |
-          if [[ -n "${CIRCLE_TAG}" ]]; then
-            wget --quiet https://github.com/gruntwork-io/fetch/releases/download/v0.3.5/fetch_linux_amd64
-            sudo install fetch_linux_amd64 /usr/bin/fetch
-            ghr_version="v0.13.0"
-            fetch --repo="https://github.com/tcnksm/ghr" --tag="${ghr_version}" --release-asset="ghr_${ghr_version}_linux_amd64.tar.gz" .
-            tar -zxvf ghr_${ghr_version}_linux_amd64.tar.gz
-            sudo install ghr_${ghr_version}_linux_amd64/ghr /usr/bin/ghr
-            which ghr
-            ./scripts/create_github_release.sh "${CIRCLE_TAG}" ./bin
-          fi
+          [[ -n "${CIRCLE_TAG}" ]] || exit 0
 
+          wget --quiet https://github.com/gruntwork-io/fetch/releases/download/v0.3.5/fetch_linux_amd64
+          sudo install fetch_linux_amd64 /usr/bin/fetch
+          ghr_version="v0.13.0"
+          fetch --repo="https://github.com/tcnksm/ghr" --tag="${ghr_version}" --release-asset="ghr_${ghr_version}_linux_amd64.tar.gz" .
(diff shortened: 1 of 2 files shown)


viswajithiii:
Created https://github.com/stackrox/kube-linter/releases/tag/test when I pushed a tag `test` on this branch.

viswajithiii:
@misberner @connorgorman  Updated to use https://github.com/tcnksm/ghr.

.circleci/config.yml
```diff
@@ -54,6 +54,20 @@ jobs:
           echo "Version from kube-linter: ${version}. Expected version: ${expected_version}"
           [[ "${version}" == "${expected_version}" ]]
 
+    - run:
```
misberner, line 57: Note that I think this should be a separate job in the workflow, but can understand if this feels overkill

scripts/create_github_release.sh
```diff
@@ -0,0 +1,24 @@
...
+bin_dir="$2"
+[[ -n "${tag}" && -n "${bin_dir}" ]] || die "Usage: $0 <tag> <binary_directory>"
+
+mkdir /tmp/release-artifacts
+for os in darwin linux windows; do
+  bin_name="kube-linter"
+  if [[ "${os}" == "windows" ]]; then
+    bin_name="kube-linter.exe"
+  fi
+  tar -C "${bin_dir}/${os}" -czf "/tmp/release-artifacts/kube-linter-${os}.tar.gz" "${bin_name}"
```
misberner, line 20: Why do we package the binary in a tar gz instead of uploading it as-is? Looks like this is common. GH restriction? Security? Purely for preserving permissions?

scripts/create_github_release.sh
```diff
@@ -0,0 +1,24 @@
...
+
+tag="$1"
+bin_dir="$2"
+[[ -n "${tag}" && -n "${bin_dir}" ]] || die "Usage: $0 <tag> <binary_directory>"
+
+mkdir /tmp/release-artifacts
+for os in darwin linux windows; do
+  bin_name="kube-linter"
+  if [[ "${os}" == "windows" ]]; then
+    bin_name="kube-linter.exe"
```
misberner, line 18: It doesn't make a lot of sense to package a windows EXE in a tgz. Maybe distribute everything as both zip and tgz?

scripts/create_github_release.sh
```diff
@@ -0,0 +1,24 @@
...
+
+set -eo pipefail
+
+[[ -n "${GITHUB_TOKEN}" ]] || die "No GITHUB_TOKEN found."
+
+tag="$1"
+bin_dir="$2"
+[[ -n "${tag}" && -n "${bin_dir}" ]] || die "Usage: $0 <tag> <binary_directory>"
+
+mkdir /tmp/release-artifacts
```
misberner, line 14: just use `mktemp -d` for this

misberner (changes requested):

.circleci/config.yml
viswajithiii: Yeah, I didn't think it was worth the cruft of persisting and attaching from workspace etc

scripts/create_github_release.sh
viswajithiii: Mostly file size, plus looking at what some other projects do. It was ~30 MB without compression.

scripts/create_github_release.sh
viswajithiii: Kubernetes puts Windows in a tar.gz: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#client-binaries But yeah, no harm in distributing everything as zip + tgz.

scripts/create_github_release.sh
viswajithiii: Done.

.circleci/config.yml
```diff
@@ -54,6 +54,20 @@ jobs:
...
+        name: Create a GitHub release, if on tag.
+        command: |
+          if [[ -n "${CIRCLE_TAG}" ]]; then
```
misberner, line 60: nit: `[[ -n "${CIRCLE_TAG}" ]] || exit 0`

misberner (approved):

PR #10, opened 2020-10-20
Refactor parameters to support arbitrary types

viswajithiii:
Instead of a `map[string]string`, support (almost) arbitrary parameters via a `map[string]interface{}`.

Use k8s-style methodology to describe the objects: ie, describe them as Go structs, embedding metadata in comments, and parse out the properties from that using the same library k8s uses, and then use code generation for the validation etc.

Also add a new check template (and built-in check) for forbidden API object versions.

## How to review this PR

The diff is smaller than it looks, because a lot of code is generated. I recommend:
1. Looking at a `params.go` file inside a template to see how the params are defined [here](https://github.com/stackrox/kube-linter/pull/10/files#diff-5f0211a31aac0aa8b344bc1148af05280cbb8f1d53806bc78a6b4b8b6a7af702).
2. Looking at the file that generates code from the `params.go` [here](https://github.com/stackrox/kube-linter/pull/10/files#diff-1f841c88832013b8502b87f9f2e1224978b63014b99b017e9befbd325b22782f). You can look at example generated code [here](https://github.com/stackrox/kube-linter/pull/10/files#diff-5f0211a31aac0aa8b344bc1148af05280cbb8f1d53806bc78a6b4b8b6a7af702).

The rest of the PR is just consequent downstream changes after doing this.


internal/templates/codegen/main.go
```diff
@@ -0,0 +1,288 @@
...
+	"golang.stackrox.io/kube-linter/internal/check"
+	"golang.stackrox.io/kube-linter/internal/templates/util"
+)
+
+var (
+	// Use this in case it doesn't get used otherwise.
+	_ = util.MustParseParameterDesc
+
+{{ range . }}
+	{{ .ParamDesc.Name}}paramDesc = util.MustParseParameterDesc({{backtick}}
```
kreamkorokke, line 56: {{ .ParamDesc.Name}}ParamDesc? to keep camel case

internal/templates/codegen/main.go
```diff
@@ -0,0 +1,288 @@
...
+{{- .ParamJSON -}}
+{{backtick}})
+{{- end }}
+
+	ParamDescs = []check.ParameterDesc{
+		{{- range . }}
+		{{ .ParamDesc.Name}}paramDesc,
```
kreamkorokke, line 63: same here

kreamkorokke (approved):
Approving since there are only minor comments

internal/pointers/pointers.go
```diff
@@ -0,0 +1,6 @@
+package pointers
+
+// Bool returns a pointer to a bool.
+func Bool(b bool) *bool {
```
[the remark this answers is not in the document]
viswajithiii: It's a minor convenience thing, which allows you to assign a value to a `*bool` in one line. We have it in the `rox` repo as well.

For example, here, `!p.NoRegex` is not addressable:

```go
	out.RegexAllowed = pointers.Bool(!p.NoRegex)
```
Without this function, I'd have to do:
```go
        regexAllowed := !p.NoRegex
	out.RegexAllowed = &regexAllowed
```

This isn't exactly a standard, but it's a pretty common pattern.

internal/stringutils/split.go
```diff
@@ -0,0 +1,16 @@
+package stringutils
+
+import (
+	"strings"
+)
+
+// Split2 splits the given string at the given separator, returning the part before and after the separator as two
+// separate return values.
+// If the string does not contain `sep`, the entire string is returned as the first return value.
+func Split2(str, sep string) (string, string) {
```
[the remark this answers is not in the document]
viswajithiii: This is useful to have when you only care about the first part of the string and little nicer to use in the caller site (and slightly more efficient, although not enough to make a difference in practice).

internal/templates/codegen/main.go
viswajithiii: Good catch, done.

internal/templates/codegen/main.go
viswajithiii: Done.

PR #12, opened 2020-10-21
First pass at Helm support

viswajithiii:
Very first pass at supporting Helm. If, when crawling directories, we find a Helm chart, we use the Helm library to render it and pass the rendered YAML files to the regular loader.

Also, refactor the loader so that there's exactly one LintContext per directory, instead of having one in total. The issue with having just one LintContext is that checks which depend on other objects become more expensive, and we risk making mistakes. (For example, if you're checking whether a service has a label that matches a pod, you may accidentally think it _does_ match a pod if it matches a pod in some other random Helm chart/directory, even if it doesn't match any pod that's actually deployed with it.)


kreamkorokke (approved):
I guess just one general comment: is array of LintContext the best way to represent it? I think as you have already mentioned in one of the TODOs, if something is useful by looking at multiple directories, I think the current design of checks need to change to accommodate this.

viswajithiii:
> I guess just one general comment: is array of LintContext the best way to represent it? I think as you have already mentioned in one of the TODOs, if something is useful by looking at multiple directories, I think the current design of checks need to change by some good amount to accommodate this.

Hmm, the idea was that each LintContext is an entirely self-contained set of objects that can be linted as a unit, so I think having an array is fine. For example, if someone runs `kube-linter lint .` in the root of a repository, there will be multiple lint contexts. The thing that's tricky is figuring out _how_ to partition the repository into contexts. For example, if people organize their YAMLs for a single app into multiple sub-directories, then we could have a case where we think a service is not referencing a valid deployment, but the deployment is just in a different folder. I don't know if that's how people do it in practice, though, so I guess we'll have to wait for user feedback for that.

PR #13, opened 2020-10-21
Add support for ignore directives

viswajithiii:
* Add support for ignoring specific checks for certain objects by adding annotations. See the [docs](https://github.com/stackrox/kube-linter/pull/13/files#diff-0b5ca119d2be595aa307d34512d9679e49186307ef94201e4b3dfa079aa89938R40) for details.
* Be more strict about applicable check names so as to make this possible.


kreamkorokke (approved):
lgtm. I think by now since we have a good amount of features built in, maybe it is a good idea to setup a test data directory to contain some test input and outputs and add an e2e test.

PR #14, opened 2020-10-23
Add remediation information to all checks

viswajithiii:
* For each check, print information about how it can be remediated.
* Add a unit test to enforce that this is added for all built-in checks.


kreamkorokke (approved):
lgtm

Issue #29, opened 2020-10-29
'make build' on clean Mac with GoLang fails with " /bin/bash: packr: command not found"


tdensmore:
Catalina 10.15.7
Golang: go version go1.15.3 darwin/amd64

~/w/r/kube-linter ❯❯❯brew install go
==> Downloading https://homebrew.bintray.com/bottles/go-1.15.3.catalina.bottle.tar.gz
==> Downloading from https://d29vzk4ow07wi7.cloudfront.net/6bd0119a31c1ae26831105490a67b97bed5d0fd2c6ad140cdf4c84aa7a315af2?response-content-disposition=attachment%3Bfilename%3D%22go-1.15.3.catalina.bottle.tar.gz%22&Policy=eyJTdGF0
######################################################################## 100.0%
==> Pouring go-1.15.3.catalina.bottle.tar.gz
🍺  /usr/local/Cellar/go/1.15.3: 9,780 files, 494.3MB
~/w/r/kube-linter ❯❯❯ make build
+ deps
go: downloading github.com/golangci/golangci-lint v1.30.0
go: downloading github.com/ghodss/yaml v1.0.0
go: downloading k8s.io/api v0.19.2
go: downloading github.com/pkg/errors v0.9.1
go: downloading k8s.io/apimachinery v0.19.2
go: downloading honnef.co/go/tools v0.0.1-2020.1.5
go: downloading k8s.io/cli-runtime v0.19.2
go: downloading github.com/stretchr/testify v1.6.1
go: downloading k8s.io/gengo v0.0.0-20200728071708-7794989d0000
go: downloading github.com/gobuffalo/packr v1.30.1
go: downloading k8s.io/client-go v0.19.2
go: downloading github.com/Masterminds/sprig/v3 v3.1.0
go: downloading github.com/davecgh/go-spew v1.1.1
go: downloading golang.org/x/text v0.3.3
go: downloading github.com/googleapis/gnostic v0.4.1
go: downloading gopkg.in/inf.v0 v0.9.1
go: downloading github.com/modern-go/reflect2 v1.0.1
go: downloading gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
go: downloading github.com/gobuffalo/envy v1.7.1
go: downloading sigs.k8s.io/kustomize v2.0.3+incompatible
go: downloading github.com/json-iterator/go v1.1.10
go: downloading github.com/mitchellh/copystructure v1.0.0
[...]

PR #30, opened 2020-10-29
Fix issue with packr install on Mac

viswajithiii:
`go install` ended up not respecting the `GOBIN` that we set on Mac, so pass it explicitly to the shell.
Should fix https://github.com/stackrox/kube-linter/issues/29


connorgorman (approved):
LG

PR #50, opened 2020-10-30
Support list objects when parsing

### LINKED ISSUE #31: how can I get pod.yaml if there is no pod.yaml on k8s server
when I find pod.yaml on my k8s master server, there is no pod.yaml.So if I can not find a pod.yaml，How can I get pod.yaml and use kube-linter to check

viswajithiii:
This makes it possible to run `kubectl get deploy -o yaml | kube-linter lint .`

Should fix #31.  

Reviewers: feel free to merge after approval.


connorgorman (approved):
Works for standalone objects, but need to think through how we can get all of the context (e.g. will complain about service accounts not existing from a pure kubectl get pod -o yaml)

PR #53, opened 2020-11-02
feat: Add check for pod not being selected using a selector

### LINKED ISSUE #38: Add check for when the selector in a deployment doesn't match the label in the pod template
This is 100% a misconfiguration, and also a very common error that happens if you ever update labels, so would be helpful to catch.

arghya88:
Fixes #38


arghya88:
@viswajithiii for some reason it's not getting registered as new template. When I do `kube-linter templates list
` it does not appear but when I do `kube-linter checks list` it appears. What am I missing?

viswajithiii:
> @viswajithiii for some reason it's not getting registered as new template. When I do `kube-linter templates list ` it does not appear but when I do `kube-linter checks list` it appears. What am I missing?

Hi @arghya88, thanks for sending this PR! What you're missing is an import in https://github.com/stackrox/kube-linter/blob/main/internal/templates/all/all.go

arghya88:
@viswajithiii got it..Now I get below error when trying to test this on a deployment yaml using `kube-linter lint deployment.yaml`

`Error: registering default check dangling-deployment: invalid check dangling-deployment: unknown object kind: Deployment
`

internal/builtinchecks/yamls/dangling-deployment.yaml
```diff
@@ -0,0 +1,7 @@
+name: "dangling-deployment"
+description: "Alert on deployments that don't have any selector matching a label in pod template"
+remediation: "Make sure your deployment's selector correctly matches a label in pod template."
+scope:
+  objectKinds:
+    - Deployment
```
viswajithiii, line 6: This needs to be `DeploymentLike` (this was causing the error you were seeing, the object kind here has to match one of the kinds registered in `internal/objectkinds`).

internal/builtinchecks/yamls/dangling-deployment.yaml
```diff
@@ -0,0 +1,7 @@
+name: "dangling-deployment"
```
viswajithiii, line 1: (Here and for the template) Rename to `mismatching-selector`, the dangling name feels less appropriate given that the selectors are defined in the same object as the template itself.

internal/builtinchecks/yamls/dangling-deployment.yaml
```diff
@@ -0,0 +1,7 @@
...
+description: "Alert on deployments that don't have any selector matching a label in pod template"
+remediation: "Make sure your deployment's selector correctly matches a label in pod template."
+scope:
+  objectKinds:
+    - Deployment
+template: "dangling-deployment"
```
viswajithiii, line 7: Add newline at end of file

internal/extract/pod_spec.go
```diff
@@ -18,6 +19,8 @@ func PodTemplateSpec(obj k8sutil.Object) (coreV1.PodTemplateSpec, bool) {
 		}, true
 	case *batchV1Beta1.CronJob:
 		return obj.Spec.JobTemplate.Spec.Template, true
+	case *appsV1.Deployment:
+		return obj.Spec.Template, true
```
viswajithiii, line 23: You don't need to do this, this is automatically handled using reflection in the `default` case below.

internal/templates/danglingdeployment/template.go
```diff
@@ -0,0 +1,59 @@
...
+	"golang.stackrox.io/kube-linter/internal/templates"
+	"golang.stackrox.io/kube-linter/internal/templates/danglingdeployment/internal/params"
+	appsV1 "k8s.io/api/apps/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Dangling Deployments",
```
viswajithiii, line 21:
```suggestion
		HumanName:   "Mismatching Selector",
```

internal/templates/danglingdeployment/template.go
```diff
@@ -0,0 +1,59 @@
...
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				deployment, ok := object.K8sObject.(*appsV1.Deployment)
+				if !ok {
+					return nil
+				}
```
viswajithiii, line 34: So, this check is applicable not just to deployments, but also to Daemonsets, Jobs etc. Can you add a new extractor for this, and implement it similar to the `PodTemplateSpec` extractor? ie, something like:
```go
func Selector(obj k8sutil.Object) (*metaV1.LabelSelector, bool) {
	switch obj := obj.(type) {
	case *batchV1Beta1.CronJob:
		return obj.Spec.JobTemplate.Spec.Selector, true
	default:
		objValue := reflect.Indirect(reflect.ValueOf(obj))
		spec := objValue.FieldByName("Spec")
		if !spec.IsValid() {
			return coreV1.PodTemplateSpec{}, false
		}
		selector := spec.FieldByName("Selector")
		if !selector.IsValid() {
			return nil, false
		}
		labelSelector, ok := template.Interface().(*metaV1.LabelSelector)
		if ok {
			return labelSelector, true
		}
	}
        return nil, false
}
```

and then you can use this selector below.
(Caveat: just typed out the code above, you may need to tweak it to get it to compile)

internal/templates/danglingdeployment/template.go
```diff
@@ -0,0 +1,59 @@
...
+				podTemplateSpec, found := extract.PodTemplateSpec(object.K8sObject)
+				if !found {
+					return nil
+				}
+				if labelSelector.Matches(labels.Set(podTemplateSpec.Labels)) {
+					// Found!
+					return nil
+				}
+				return []diagnostic.Diagnostic{{Message: fmt.Sprintf("no pods found matching deployment labels (%v)", selector)}}
```
viswajithiii, line 55:
```suggestion
				return []diagnostic.Diagnostic{{Message: fmt.Sprintf("labels in pod spec (%v) do not match labels in selector (%v)", podTemplateSpec.Labels, selector)}}
```

internal/templates/danglingdeployment/template.go
```diff
@@ -0,0 +1,59 @@
...
+	"golang.stackrox.io/kube-linter/internal/templates/danglingdeployment/internal/params"
+	appsV1 "k8s.io/api/apps/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Dangling Deployments",
+		Key:         "dangling-deployment",
```
viswajithiii, line 22: `mismatching-selector`

internal/builtinchecks/yamls/dangling-deployment.yaml
arghya88: Done

internal/extract/pod_spec.go
arghya88: Done

internal/builtinchecks/yamls/dangling-deployment.yaml
arghya88: Done

internal/templates/danglingdeployment/template.go
arghya88: Done

internal/templates/danglingdeployment/template.go
arghya88: Done

internal/templates/danglingdeployment/template.go
arghya88: Done

internal/templates/danglingdeployment/template.go
arghya88: Done

internal/builtinchecks/yamls/dangling-deployment.yaml
arghya88: Done

arghya88:
@viswajithiii have done the suggested changes..PTAL

internal/builtinchecks/yamls/mismatching-selector.yaml
```diff
@@ -0,0 +1,7 @@
+name: "mismatching-selector"
+description: "Alert on deployments that don't have any selector matching a label in pod template"
```
viswajithiii, line 2:
```suggestion
description: "Alert on deployments where the selector doesn't match the pod template labels"
```

internal/builtinchecks/yamls/mismatching-selector.yaml
```diff
@@ -0,0 +1,7 @@
...
+remediation: "Make sure your deployment's selector correctly matches a label in pod template."
```
viswajithiii, line 3:
```suggestion
remediation: "Make sure your deployment's selector correctly matches the labels in its pod template."
```

internal/templates/mismatchingselector/template.go
```diff
@@ -0,0 +1,55 @@
...
+				if !hasSelector {
+					return []diagnostic.Diagnostic{{
+						Message: "Object has no selector specified",
+					}}
+				}
+				podTemplateSpec, hasPods := extract.PodTemplateSpec(object.K8sObject)
+				if !hasPods {
+					return nil
+				}
+				labelSelector, err := metaV1.LabelSelectorAsSelector(&metaV1.LabelSelector{MatchLabels: selector.MatchLabels, MatchExpressions: selector.MatchExpressions})
```
viswajithiii, line 41:
```suggestion
				labelSelector, err := metaV1.LabelSelectorAsSelector(selector)
```

does this not work?

internal/extract/pod_spec.go
```diff
@@ -47,3 +49,31 @@ func PodSpec(obj k8sutil.Object) (coreV1.PodSpec, bool) {
 	}
 	return podTemplateSpec.Spec, true
 }
+
+// Selector extracts a selector from the given object, if available.
+func Selector(obj k8sutil.Object) (*metaV1.LabelSelector, bool) {
```
viswajithiii, line 54: So, the second return value here is supposed to indicate whether the object is one that _has_ a selector (for example, it should return `false` if the object is a plain `Pod`). So I would remove the `hasSelector` thing from here, and just `return true` in every case where you're not explicitly returning `nil`.

internal/templates/mismatchingselector/template.go
```diff
@@ -0,0 +1,55 @@
...
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				selector, hasSelector := extract.Selector(object.K8sObject)
+
+				if !hasSelector {
+					return []diagnostic.Diagnostic{{
+						Message: "Object has no selector specified",
+					}}
+				}
```
viswajithiii, line 36: After implementing the above suggestion, rename the second return value to `found`. `if !found`, you can return `nil`, instead of an error, since that just means this object is an object without a selector (I think the only such object under `DeploymentLike` is `Pod`).  _Then_, ensure that the selector is specified using the logic from `hasSelector` above.

internal/templates/mismatchingselector/template.go
```diff
@@ -0,0 +1,55 @@
...
+				podTemplateSpec, hasPods := extract.PodTemplateSpec(object.K8sObject)
+				if !hasPods {
+					return nil
+				}
+				labelSelector, err := metaV1.LabelSelectorAsSelector(&metaV1.LabelSelector{MatchLabels: selector.MatchLabels, MatchExpressions: selector.MatchExpressions})
+				if err != nil {
+					return []diagnostic.Diagnostic{{
+						Message: fmt.Sprintf("Object has invalid label selector: %v", err),
```
viswajithiii, line 44:
```suggestion
						Message: fmt.Sprintf("object has invalid label selector: %v", err),
```

We're trying to keep diagnostic messages starting with lower-case.

internal/templates/mismatchingselector/template.go
```diff
@@ -0,0 +1,55 @@
...
+					}}
+				}
+				if labelSelector.Matches(labels.Set(podTemplateSpec.Labels)) {
+					// Found!
```
viswajithiii, line 48: Remove this comment?

viswajithiii (commented):
Thanks for working through the changes. A few more minor updates, and this should be good to go! 

After you make the changes, can you also paste an example YAML file that you tested on, and the results of running `kube-linter` on it? (I'm working on adding an E2E test framework for this, but in the interim, I want to make sure to sanity-check it manually)

arghya88:
Deployment.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx1
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```

Error: 


```
../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) labels in pod spec (map[app:nginx]) do not match labels in selector (&LabelSelector{MatchLabels:map[string]string{app: nginx1,},MatchExpressions:[]LabelSelectorRequirement{},}) (check: mismatching-selector, remediation: Make sure your deployment's selector correctly matches the labels in its pod template.)

../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) container "nginx" does not have a read-only root file system (check: no-read-only-root-fs, remediation: Set readOnlyRootFilesystem to true in your container's securityContext.)

../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) container "nginx" is not set to runAsNonRoot (check: run-as-non-root, remediation: Set runAsUser to a non-zero number, and runAsNonRoot to true, in your pod or container securityContext. See https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for more details.)

../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) container "nginx" has cpu request 0 (check: unset-cpu-requirements, remediation: Set your container's CPU requests and limits depending on its requirements. See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for more details.)

../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) container "nginx" has cpu limit 0 (check: unset-cpu-requirements, remediation: Set your container's CPU requests and limits depending on its requirements. See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for more details.)

../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) container "nginx" has memory request 0 (check: unset-memory-requirements, remediation: Set your container's memory requests and limits depending on its requirements. See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for more details.)

../../deployment.yaml: (object: <no namespace>/nginx-deployment apps/v1, Kind=Deployment) container "nginx" has memory limit 0 (check: unset-memory-requirements, remediation: Set your container's memory requests and limits depending on its requirements. See https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for more details.)

Error: found 7 lint errors
```

internal/templates/mismatchingselector/template.go
```diff
@@ -0,0 +1,58 @@
...
+	"golang.stackrox.io/kube-linter/internal/templates/mismatchingselector/internal/params"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Mismatching Selector",
+		Key:         "mismatching-selector",
+		Description: "Flag deployments which do not have any selector matching a label in pod template",
```
viswajithiii, line 22:
```suggestion
		Description: "Flag deployments where the selector doesn't match the labels in the pod template spec",
```

internal/templates/mismatchingselector/template.go
```diff
@@ -0,0 +1,58 @@
+package mismatchingselector
+
+import (
+	"fmt"
+
+	"golang.stackrox.io/kube-linter/internal/extract"
+
```
viswajithiii, line 7: Remove newline.

viswajithiii (approved):
Looks good, thanks again for your contribution! I just have these two minor comments. If you could just fix those, I'll merge this in right after. 🙂

internal/templates/mismatchingselector/template.go
arghya88: Done

internal/templates/mismatchingselector/template.go
arghya88: Done

arghya88:
@viswajithiii I just ran `golangci-lint` in makefile

viswajithiii:
Thanks for the contribution, and for working through the review comments, @arghya88! Appreciate it.

PR #54, opened 2020-11-06
feat: Add check for objects with multiple replicas not having anti-affinity set

### LINKED ISSUE #39: Add check enforcing anti-affinity for deployments with multiple replicas
Check that deployments with multiple replicas have pods set to have anti-affinity, to tell Kubernetes to try to run the pods on different nodes. Ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

arghya88:
Fixes #39


internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,45 @@
...
+					return nil
+				}
+				if *replicas > 1 {
+					podTemplateSpec, hasPods := extract.PodTemplateSpec(object.K8sObject)
+					if !hasPods {
+						return nil
+					}
+					if podTemplateSpec.Spec.Affinity != nil && podTemplateSpec.Spec.Affinity.PodAntiAffinity != nil {
+						return nil
+					}
```
viswajithiii, line 38: So what I had in mind for this check was something even more strict: specify that intra-pod anti-affinity is set for pods of this deployment, to ensure that they would be scheduled on different replicas.

Mechanically, this means ensuring that at least one of the terms in the `podAntiAffinity` (whether it's `required` or `prefered`) satisfies _both_ of the following properties:
1. topologyKey is `kubernetes.io/hostname` (so that anti-affinity is based on nodes)
2. the labelSelector matches the labels in the pod template spec (that is, it would match pods generated from the current spec)

See the following for an example (copied from [here](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node)) of a deployment that _would_ pass this check.
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: redis-cache
spec:
  selector:
    matchLabels:
      app: store
  replicas: 3
  template:
    metadata:
      labels:
        app: store
    spec:
      affinity:
        podAntiAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values:
                - store
            topologyKey: "kubernetes.io/hostname"
      containers:
      - name: redis-server
        image: redis:3.2-alpine

```

internal/templates/antiaffinity/internal/params/params.go
```diff
@@ -0,0 +1,5 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
```
viswajithiii, line 4: Can you add add `MinReplicas` as a parameter to this template, so that the users can specify a minimum number of replicas for which they want to enforce this? We can then specify the value as 2 explicitly in the built-in check.

viswajithiii (commented):
Sorry for the late review, and thanks for sending the PR. See comments.

internal/templates/antiaffinity/internal/params/params.go
arghya88: Added..but not sure how to use it.

internal/builtinchecks/yamls/no-anti-affinity.yaml
```diff
@@ -0,0 +1,11 @@
+name: "no-anti-affinity"
+description: "Alert on deployments with multiple replicas that don't specify inter pod anti-affinity to ensure that the orchestrator attempts to schedule replicas on different nodes"
+remediation: >-
+  Specify anti-affinity in your pod spec to ensure that the orchestrators attempts to schedule replicas on different nodes.
+  You can do this by using podAntiAffinity, specifying a labelSelector that matches pods of this deployment,
+  and setting the topologyKey to kubernetes.io/hostname
```
viswajithiii, line 6:
```suggestion
  and setting the topologyKey to kubernetes.io/hostname.
```

internal/templates/antiaffinity/internal/params/params.go
```diff
@@ -0,0 +1,8 @@
...
+
+	//Minimum number of replicas
```
viswajithiii, line 6:
```suggestion
	// The minimum number of replicas a deployment must have before anti-affinity is enforced on it
```

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,71 @@
...
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				replicas, found := extract.Replicas(object.K8sObject)
+				if !found {
+					return nil
+				}
+				if replicas > 1 {
```
viswajithiii, line 34:
```suggestion
				if int(replicas) >= p.MinReplicas {
```

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,71 @@
...
+	templates.Register(check.Template{
+		HumanName:   "Anti affinity not specified",
+		Key:         "anti-affinity",
+		Description: "Flag objects with multiple replicas but anti affinity not specified in the pod template spec",
+		SupportedObjectKinds: check.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
```
viswajithiii, line 28:
```suggestion
		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
```

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,71 @@
...
+				replicas, found := extract.Replicas(object.K8sObject)
+				if !found {
+					return nil
+				}
+				if replicas > 1 {
+					podTemplateSpec, hasPods := extract.PodTemplateSpec(object.K8sObject)
+					if !hasPods {
+						return nil
+					}
+					if podTemplateSpec.Spec.Affinity != nil && podTemplateSpec.Spec.Affinity.PodAntiAffinity != nil {
```
viswajithiii, line 39:
```suggestion
					if affinity := podTemplateSpec.Spec.Affinity; affinity != nil && affinity.PodAntiAffinity != nil {						
```

(and use `affinity` everywhere inside)

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,71 @@
...
+						preferredAffinity := podTemplateSpec.Spec.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
+						requiredAffinity := podTemplateSpec.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+						for _, preferred := range preferredAffinity {
+							labelSelector, err := metaV1.LabelSelectorAsSelector(preferred.PodAffinityTerm.LabelSelector)
+							if err != nil {
+								return []diagnostic.Diagnostic{{
```
viswajithiii, line 45: Instead of returning, let's just ignore these and continue. We need to add separate checks to validate objects in general, but I would prefer to keep this check focused on this.

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,71 @@
...
+						for _, required := range requiredAffinity {
+							labelSelector, err := metaV1.LabelSelectorAsSelector(required.LabelSelector)
+							if err != nil {
+								return []diagnostic.Diagnostic{{
+									Message: fmt.Sprintf("pod anti-affinity has invalid label selector: %v", err),
+								}}
+							}
+							if required.TopologyKey == "kubernetes.io/hostname" && labelSelector.Matches(labels.Set(podTemplateSpec.Labels)) {
+								return nil
+							}
```
viswajithiii, line 62: Extract this component into a function. The signature can be:
```go
func(PodAffinityTerm) bool
```

internal/builtinchecks/yamls/no-anti-affinity.yaml
```diff
@@ -0,0 +1,11 @@
...
+  See https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity for more details.
+scope:
+  objectKinds:
+    - DeploymentLike
+template: "anti-affinity"
```
viswajithiii, line 11:
```suggestion
template: "anti-affinity"
params:
  minReplicas: 2
```

arghya88:
@viswajithiii Addressed all the comments.PTAL

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,68 @@
...
+					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("object with (%v) replicas but anti affinity is not specified in pod spec ", replicas)}}
+				}
+				return nil
+			}, nil
+		}),
+	})
+}
+
+func isAffinityTerm(affinityTerm coreV1.PodAffinityTerm, podTemplateSpec coreV1.PodTemplateSpec) bool {
+	labelSelector, _ := metaV1.LabelSelectorAsSelector(affinityTerm.LabelSelector)
```
viswajithiii, line 63: You still need to check the error and `return false` if `err != nil`. Otherwise it will be `nil` and the next line will panic.

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,68 @@
...
+					}
+					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("object with (%v) replicas but anti affinity is not specified in pod spec ", replicas)}}
+				}
+				return nil
+			}, nil
+		}),
+	})
+}
+
+func isAffinityTerm(affinityTerm coreV1.PodAffinityTerm, podTemplateSpec coreV1.PodTemplateSpec) bool {
```
viswajithiii, line 62: Just pass the labels in the second argument.

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,68 @@
...
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			return func(_ *lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				replicas, found := extract.Replicas(object.K8sObject)
+				if !found {
+					return nil
+				}
+				if int(replicas) >= p.MinReplicas {
```
viswajithiii, line 35: To reduce indentation levels, can you do:
```go
if int(replicas) < p.MinReplicas {
  return nil
}
```
and then put the rest?

internal/templates/antiaffinity/template.go
viswajithiii, line 62: Rename function to `affinityTermMatchesLabelsAgainstNodes`

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,68 @@
...
+								return nil
+							}
+						}
+						for _, required := range requiredAffinity {
+							if isAffinityTerm(required, podTemplateSpec) {
+								return nil
+							}
+						}
+					}
+					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("object with (%v) replicas but anti affinity is not specified in pod spec ", replicas)}}
```
viswajithiii, line 54:
```suggestion
					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("object has %d replicas but does not specify inter pod anti-affinity", replicas)}}
```

internal/templates/antiaffinity/template.go
```diff
@@ -0,0 +1,68 @@
...
+	"golang.stackrox.io/kube-linter/internal/templates"
+	"golang.stackrox.io/kube-linter/internal/templates/antiaffinity/internal/params"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Anti affinity not specified",
+		Key:         "anti-affinity",
+		Description: "Flag objects with multiple replicas but anti affinity not specified in the pod template spec",
```
viswajithiii, line 23:
```suggestion
		Description: "Flag objects with multiple replicas but inter-pod anti affinity not specified in the pod template spec",
```

internal/templates/antiaffinity/internal/params/params.go
```diff
@@ -0,0 +1,8 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
+
+	//The minimum number of replicas a deployment must have before anti-affinity is enforced on it
```
viswajithiii, line 6:
```suggestion
	// The minimum number of replicas a deployment must have before anti-affinity is enforced on it
```

viswajithiii (approved):
LGTM, after few minor changes. I'll merge whenever you're ready. Thanks for the contribution!

PR #55, opened 2020-11-06
style(checks): fix typo

jai:


viswajithiii:
Hi @jai, thanks for the PR! Note that this doc is generated from the code in https://github.com/stackrox/kube-linter/blob/main/internal/builtinchecks/yamls/deprecated-service-account.yaml#L3. So you will have to update that, and then run `make generated-docs` to generate the updated docs (which should match the diff you have already made manually).

jai:
Ah will do thanks

viswajithiii:
Thanks for the contribution @jai! I've gone ahead and made the fix.

Issue #23, opened 2020-10-29
Docker image for kube-linter


linuxbsdfreak:
Hi,

It would be nice if there would be a official docker image for kube-linter. Which would help in usage in CI/CD process .

Kevin

viswajithiii:
Hi @linuxbsdfreak,

Thanks for filing this issue! This is definitely something we can do, but first I'd like to understand how you would like to use it to make sure we can handle that. Specifically, how do you want to pass the files to be linted to KubeLinter? Is your goal to be able to do something like `docker run -v <path-to-helm-files>:<path-in-container> <kube-linter-image>`?

tdensmore:
That would be my use case... to avoid installing it locally, and allow it to be run from CI/CD tooling.

linuxbsdfreak:
Hi @viswajithiii ,

Thanks. I am using tekton CI/CD to run tools to validate helm charts. I would like to use  kube-linter  as a tool packaged as a docker image  and run it against helm charts path or plain K8s manifests. That would be my use case. 

Kevin

kreamkorokke:
Hi @linuxbsdfreak and @tdensmore 
That makes sense. We have added it to our roadmap and thanks again for bringing this up. I will loop this back to our main developer.

jspiro:
Has anyone requested official support for running this using GitHub Actions? I can't find any such request. That would be our use-case, to run the above image in PRs using GitHub natively.

viswajithiii:
> Has anyone requested official support for running this using GitHub Actions? I can't find any such request. That would be our use-case, to run the above image in PRs using GitHub natively.

Hi @jspiro, yes, we have discussed having an official GitHub Action before, but haven't tracked it in an issue as yet. I've created #96 to track this.

Even before that, though, it _should_ be possible to use KubeLinter without too much difficulty in a GitHub Action. We have an example [here](https://github.com/stackrox/contributions/tree/master/ci/GitHub/kube-linter) that may be helpful (it uses the binary directly instead of the image, which I think is preferable anyway since it avoids the dependency on Docker).

jspiro:
Aw, thank you @viswajithiii! I will give the example a shot and otherwise follow #96. Maybe I'll even find some time to contribute to it ;-)

PR #68, opened 2020-11-11
Add Dockerfile and GitHub action for pushing

crunchtime-ali:
Here is a `Dockerfile` with a multi-stage build that uses `FROM:scratch` in the final image. Therefore, it only weighs 30.8 MB (compared to 751.0 MB in the first stage).

I also added two GitHub actions:
`docker-image.yml` builds the image on every release and pushes it on the DockerHub to `stackrox/kube-linter`. You will need to get a username and token from your `stackrox` organization on DockerHub and then save those in the GitHub project settings as `DOCKER_USERNAME` and `DOCKER_PASSWORD`
`linter.yml` uses GitHub's SuperLinter which itself uses multiple Linter's. For this project it should use:
- hadolint / dockerfilelint for the Dockerfile
- golangci-lint for the Go code

Originally I only had a `hadolint` action but using the SuperLinter seems to be the more convenient option. If you want I can also disable `golangci-lint` but I am eager to hear your opinion on it first.

This fixes https://github.com/stackrox/kube-linter/issues/23


viswajithiii:
@crunchtime-ali Thanks for the PR! I'm going to push a couple of commits to update the Dockerfile and the env variable name to this branch before merging.

viswajithiii:
Thanks again for the contributions, @crunchtime-ali! For simplicity with our existing tooling, I've decided to just stick with CircleCI for now versus adding GitHub actions, although I do want to migrate eventually since it looks pretty nice. I also changed the Dockerfile slightly so as to just reuse the code for `build` from the Makefile.

I've created the DockerHub repository [here](https://hub.docker.com/r/stackrox/kube-linter).

viswajithiii (approved):

PR #71, opened 2020-11-12
Store built binaries as artifacts in CI

viswajithiii:
This allows for easier testing on PRs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -91,6 +91,9 @@ jobs:
           which ghr
           ./scripts/create_github_release.sh "${CIRCLE_TAG}" ./bin
 
+    - store_artifacts:
+        path: bin
+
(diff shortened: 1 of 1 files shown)


kreamkorokke (approved):

kreamkorokke:
Actually sorry why do we need to store it in bin? I thought each PR will build its own binaries?

viswajithiii:
@kreamkorokke This basically takes the binaries and stores them as an artifact in CircleCI, so that we can easily download them. It's useful if you want to test the KubeLinter binary corresponding to a PR without building it. See https://app.circleci.com/pipelines/github/stackrox/kube-linter/236/workflows/7d218b78-bf58-42f8-9723-d115379d94db/jobs/606/artifacts for the examples from this PR.

PR #74, opened 2020-11-13
Add sanity test to ensure KubeLinter doesn't crash on any charts from the Helm chart repo

viswajithiii:
I found a bug where we would crash on CronJobs and Jobs without selectors defined. Fix this bug, and also add an E2E test that pulls the Helm chart repo from https://github.com/helm/charts and runs KubeLinter on it, and makes sure it doesn't panic.

Example build showing failure before I fixed the issue: https://app.circleci.com/pipelines/github/stackrox/kube-linter/250/workflows/c7cf15e6-4d3b-43ef-badd-92c1575ce77a/jobs/656

### DIFF
--- a/e2etests/empty.go
+++ b/e2etests/empty.go
@@ -1,0 +1,3 @@
+package e2etests
+
+// Empty file with no build tag to keep the Go compiler happy.
--- a/e2etests/sanity_test.go
+++ b/e2etests/sanity_test.go
@@ -1,0 +1,44 @@
+// +build e2e
+
+package e2etests
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	kubeLinterBinEnv = "KUBE_LINTER_BIN"
+)
+
+func TestKubeLinterWithBuiltInChecksDoesntCrashOnHelmChartsRepo(t *testing.T) {
+	kubeLinterBin := os.Getenv(kubeLinterBinEnv)
+	require.NotEmpty(t, kubeLinterBin, "Please set %s", kubeLinterBinEnv)
+
+	_, err := os.Stat(kubeLinterBin)
+	require.NoError(t, err)
+
+	tmpDir, err := ioutil.TempDir("", "")
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, os.RemoveAll(tmpDir))
+	}()
+
+	chartsDir := filepath.Join(tmpDir, "charts")
+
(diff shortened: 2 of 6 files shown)


e2etests/sanity_test.go
```diff
@@ -0,0 +1,44 @@
...
+		require.NoError(t, os.RemoveAll(tmpDir))
+	}()
+
+	chartsDir := filepath.Join(tmpDir, "charts")
+
+	gitCloneOut, err := exec.Command("git", "clone", "git@github.com:helm/charts.git", chartsDir).CombinedOutput()
+	require.NoError(t, err, "Git clone failed. output: %s", string(gitCloneOut))
+
+	kubeLinterOut, err := exec.Command(kubeLinterBin, "lint", chartsDir, "--config", "testdata/all-built-in-config.yaml").CombinedOutput()
+	// Something will fail for sure, so kube-linter will not return a success code.
```
kreamkorokke, line 39: I think this is definitely a good start but should we consider removing this dependency to helm's repo? So that we can further verify the specific checks that failed.

e2etests/empty.go
```diff
@@ -0,0 +1,3 @@
+package e2etests
+
+// Empty file with no build tag to keep the Go compiler happy.
```
kreamkorokke, line 3: Not a PR comment but why does go require an empty file for compilation?

e2etests/sanity_test.go
viswajithiii: The reason I like this repo is that it has a very wide variety of charts with all sorts of configurations. This test is literally just to make sure that KubeLinter: a) finds _some_ lint errors in the charts and b) doesn't panic.

We should definitely add tests that take specific YAML files and verify the specific checks that failed. I'm working on that too, but this test will still be relevant because we get a huge amount of breadth for free from the Helm repo

e2etests/empty.go
viswajithiii: Basically, some go commands require that there is at least one file in each package. Since the other file in this package has a build tag (that's the `+build e2e` line on top), it is not included by default unless the go command passes `-tags e2e`, and so go complains the package is empty.

(Build tags are just a useful way to selectively include files under certain conditions (for example, if you need to have a different implementation depending on the OS, you can do that). In this case, I gave all e2etests a build tag of `e2e` so that they don't get run if someone simply runs `go test ./...` to run unit tests, they get run only if someone explicitly asks for e2es.

Ref: https://golang.org/pkg/go/build/#hdr-Build_Constraints)

kreamkorokke (approved):

e2etests/empty.go
kreamkorokke: gotcha. thanks!

e2etests/sanity_test.go
kreamkorokke: makes sense

PR #100, opened 2020-12-02
chore: links to contribution and code of conduct doc

nirav24:


viswajithiii (approved):
Thanks for the catch!

PR #104, opened 2020-12-04
Add capabilities check

### LINKED ISSUE #32: Add checks for capabilities
Allow users to write checks to enforce that containers are dropping capabilities, or not adding capabilities. Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container.

kreamkorokke:
This resolves #32


internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil, nil
+}
+
+func checkRequiredDrops(containerName string, paramCaps []string, scCaps []v1.Capability) (*diagnostic.Diagnostic, error) {
+	for _, requiredCap := range paramCaps {
+		requiredCapMatcher, err := matcher.ForString(requiredCap)
+		if err != nil {
+			return nil, err
+		}
+		found := false
```
viswajithiii, line 46:
```suggestion
		var found bool
```

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,107 @@
...
+		HumanName:   "Verify container capabilities",
+		Key:         "verify-container-capabilities",
+		Description: "Flag containers that do not match capabilities requirements",
+		SupportedObjectKinds: check.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			var returnedErr error
```
viswajithiii, line 80: There's no point modifying this error inside the `check.Func`, since this error is checked long before any `check.Func` is invoked. We should do everything that could potentially result in an error (like constructing matchers from the params) _outside_ the closure, and `return nil, error` here in that case.

internal/templates/containercapabilities/internal/params/params.go
```diff
@@ -0,0 +1,13 @@
...
+// Params represents the params accepted by this template.
+type Params struct {
+
+	// List of capabilities that are required to be dropped by containers.
+	// +notnegatable
+	RequiredDrops []string `json:"requiredDrops"`
+
+	// List of capabilities that are forbidden to be added to containers.
+	// +notnegatable
+	ForbiddenAdds []string `json:"forbiddenAdds"`
```
viswajithiii, line 12: As discussed offline, handle the `all` case and also account for the defaults.

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	reservedCapabilitiesAll             = "all"
+	matchLiteralReservedCapabilitiesAll = "(?i)" + reservedCapabilitiesAll
+)
+
+var (
+	literalReservedCapabilitiesAllMatcher func(string) bool
```
viswajithiii, line 23: This can just be initialized statically:
```go
	literalReservedCapabilitiesAllMatcher = func() func(string) bool {
               m, err := matcher.ForString(matchLiteralReservedCapabilitiesAll) 
               utils.Must(err)
               return m
        }()

```

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			var err error
+			literalReservedCapabilitiesAllMatcher, err = matcher.ForString(matchLiteralReservedCapabilitiesAll)
+			if err != nil {
+				return nil, err
+			}
+
+			// Check for "all" in case user does not want any capability in containers.
+			forbidAll := checkForbidAll(p.ForbiddenCapabilities)
```
viswajithiii, line 165: Should we also enforce that if `all` is in the list, then the list has no other elements?

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+func checkForbidAll(paramCaps []string) bool {
+	for _, paramCap := range paramCaps {
+		if literalReservedCapabilitiesAllMatcher(paramCap) {
+			return true
+		}
+	}
+	return false
+}
+
+func verifyExceptionsList(forbidAll bool, exceptions []string) error {
```
viswajithiii, line 132: Call this `validate`?

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+					if err != nil {
+						return nil, errors.Wrapf(err, "checking container capabilities. invalid capability: %s", cap)
+					}
+					exceptionCapMatchers[cap] = capMatcher
+				}
+			}
+
+			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
+				// 2 = 1 for forbiddenAdds + 1 for allRequiredDrops
+				result := make([]diagnostic.Diagnostic, 0, 2)
```
viswajithiii, line 193: Just do `var result []diagnostic.Diagnostic`, since the slice may be empty

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+					containerName,
+					scCaps.Drop,
+					paramCap),
+			}
+		}
+	}
+	return nil
+}
+
+func checkCapabilityAddList(
```
viswajithiii, line 73: You can make this function take in a pointer to the output slice (ie, a `*[]diagnostic.Diagnostic`) and just append your findings to the slice.

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+			var excluded bool
+			for _, exceptionCapMatcher := range exceptionCapMatchers {
+				if exceptionCapMatcher(string(scCap)) {
+					// Forgive this capability
+					excluded = true
+					break
+				}
+			}
+			if !excluded {
+				return &diagnostic.Diagnostic{
```
viswajithiii, line 92: We should have one message for each forbidden capability.

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+				}
+			}
+		}
+		// No violations
+		return nil
+	}
+
+	// Any capability from scCaps should not match with any from paramCaps
+	for _, paramCapMatcher := range paramCapMatchers {
+		for _, scCap := range scCaps.Add {
```
viswajithiii, line 107: So, the tricky thing here is that containers can have some capabilities implicitly, through defaults. This means we should either:
1. Use the default list of defaults from [here](https://github.com/moby/moby/blob/master/oci/caps/defaults.go#L6-L19), and assume they are in the container.
1. Be extra-restrictive, and say that to forbid a capability, you _must_ drop it explicitly. (This is the same as assuming that containers have all capabilities by default.)

As written, the `NET_RAW` check will not fire by default on an empty capabilities list. But it should.

internal/templates/containercapabilities/template.go
```diff
@@ -0,0 +1,218 @@
...
+				scCaps.Drop),
+		}
+	}
+
+	// Every forbidden capability specified by param should be found in the DROP list
+	for paramCap, paramCapMatcher := range paramCapMatchers {
+		var found bool
+		for _, scCap := range scCaps.Drop {
+			// User can specify to drop "all" under containers as well
+			if paramCapMatcher(string(scCap)) || literalReservedCapabilitiesAllMatcher(string(scCap)) {
```
viswajithiii, line 54: In general, I think we should just compute the effective list of capabilities by looking at the "add" and "drop", and pass those into this function.

internal/templates/containercapabilities/template.go
kreamkorokke: Yes it will not fire on checking ADD list, but it still fires on checking DROP list right?

internal/templates/containercapabilities/template.go
kreamkorokke: yeah I thought about it as well, but that requires keeping a base list that is consistent with the default caps, which I was not sure if it changes or not.

The cleanest way of doing this would be:
- 1: find a base list
- 2: perform add and drop
- 3: validate against the user specified params.

However, since we don't have that base list, and not sure how safe it is to just hard code one (not only this can changes, this has to also work with different versions of kubernetes server), I instead chose to check it by making sure unwanted caps are:
- 1: not found in ADD list since this overwrites even if it is included in DROP list
- 2: dropped in DROP list

viswajithiii (approved):
Looks good!

PR #105, opened 2020-12-04
Add support for slice type in params

kreamkorokke:


internal/templates/codegen/main.go
```diff
@@ -205,6 +205,8 @@ func constructParameterDescsFromStruct(typeSpec *types.Type) ([]check.ParameterD
 			default:
 				return nil, errors.Errorf("currently unsupported type %v", member.Type)
 			}
+		case types.Slice:
```
viswajithiii, line 208: This is necessary, but not sufficient. For example, we need to figure out how we'd represent array parameters in the generated template docs (https://github.com/stackrox/kube-linter/blob/main/docs/generated/templates.md). And we need to way to describe what type the elements of the array should be for that.

I suggest doing this as part of #104, since you actually use it there and you can actually make sure you handle the different use-cases.

kreamkorokke: Yeah I am actually in the midst of testing this with the other PR. Sure I can merge them into one then

PR #106, opened 2020-12-07
Fix formatting for service-account regex

mavimo:
Markdown `|` need to be escaped


docs/generated/checks.md
```diff
@@ -5,7 +5,7 @@ KubeLinter includes the following built-in checks:
 | Name | Enabled by default | Description | Remediation | Template | Parameters |
 | ---- | ------------------ | ----------- | ----------- | -------- | ---------- |
  | dangling-service | Yes | Alert on services that don't have any matching deployments | Make sure your service's selector correctly matches the labels on one of your deployments. | dangling-service | `{}` |
- | default-service-account | No | Alert on pods that use the default service account | Create a dedicated service account for your pod. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/ for more details. | service-account | `{"serviceAccount":"^(|default)$"}` |
+ | default-service-account | No | Alert on pods that use the default service account | Create a dedicated service account for your pod. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/ for more details. | service-account | `{"serviceAccount":"^(\|default)$"}` |
```
kreamkorokke, line 8: Hi @mavimo, first of all thanks for submitting a PR. 

The file you modified is actually a generated file, from [this yaml file](https://github.com/stackrox/kube-linter/blob/main/internal/builtinchecks/yamls/default-service-account.yaml), via this [command](https://github.com/stackrox/kube-linter/blob/main/internal/command/checks/command.go). Please instead modify the source files to achieve this. Thanks!

kreamkorokke: Also just to add to the above, since changing the source code involves more than just one line change, if you are unfamiliar with golang or does not have the time, please feel free to let us know and we can do it as well:)

mavimo: @kreamkorokke thx for your feedback, i'll try to fix it asap ;)

mavimo: @kreamkorokke I try to solve via https://github.com/stackrox/kube-linter/pull/107 let me know if the current approach is fine or you prefer just to have a single filter function that escape the pipe.

I'm going to close this PR.

Issue #78, opened 2020-11-13
Allow all options in the config file to alternatively be specified as command-line flags


viswajithiii:
The one exception to this would be custom checks.

Perhaps using https://github.com/spf13/viper.

PR #109, opened 2020-12-11
ADD: support config options as CLI flags

nirav24:
fixes https://github.com/stackrox/kube-linter/issues/78

@viswajithiii Could you please review and provide feedback/suggestions?


internal/command/lint/command.go
```diff
@@ -89,7 +91,19 @@ func Command() *cobra.Command {
...
 		},
 	}
-	c.Flags().StringVar(&configPath, "config", "", "Path to config file")
+
+	c.Flags().String("config", "", "Path to config file")
 	c.Flags().BoolVarP(&verbose, "verbose", "v", false, "Enable verbose logging")
+	c.Flags().StringSlice("exclude", nil, "Exclude checks")
+	c.Flags().StringSlice("include", nil, "Include provided checks")
+	c.Flags().Bool("add-all-built-in", false, "Add All BuildIn checks")
+	c.Flags().Bool("no-defaults", false, "Don't add Default checks")
```
viswajithiii, line 100: We should make these names match the ones in the config (but convert camelCase to dashed-case).

The fact that we need to repeat the config parameters between Viper and the flags is not ideal. I think we can use code generation (similar to what we have done for our [templates](https://github.com/stackrox/kube-linter/blob/main/internal/templates/codegen/main.go)) where we generate code that can populate these flags automatically. That way, there is only one source of truth.

To be concrete, we should be able to write code that looks at:
```go
// ChecksConfig is the config that determines which checks to run.
type ChecksConfig struct {
	// AddAllBuiltIn, if set, adds all built-in checks. This allows users to
	// explicitly opt-out of checks that are not relevant using Exclude.
	AddAllBuiltIn bool `json:"addAllBuiltIn"`
	// DoNotAutoAddDefaults, if set, prevents the automatic addition of default checks.
	DoNotAutoAddDefaults bool `json:"doNotAutoAddDefaults"`
	// Exclude is a list of check names to exclude.
	Exclude []string `json:"exclude"`
	// Include is a list of check names to include. If a check is in both Include and Exclude,
	// Exclude wins.
	Include []string `json:"include"`
}

// Config represents the config file format.
type Config struct {
        // +viper=exclude
	CustomChecks []check.Check `json:"customChecks,omitempty"`
	Checks       ChecksConfig  `json:"checks,omitempty"`
}
```
walks the struct, relies on the comments for descriptions, and generates a function that can add flags for all paths that do NOT have a `viper=exclude` comment, like:
```
func AddFlags(c *cobra.Command, v *viper.Viper) {
    c.Flags().StringSlice("exclude", nil, "Exclude is a list of check names to exclude.")
     v.BindPFlag("checks.exclude", c.Flags().Lookup("exclude")
     // ... and so on
}
```

viswajithiii (commented):
Left a few comments, but thanks for getting a start on this!

internal/config/config.go
```diff
@@ -29,14 +30,26 @@ type Config struct {
...
+		if err := v.ReadInConfig(); err != nil {
+			return Config{}, errors.Wrap(err, "reading file")
+		}
 	}
-	var c Config
-	if err := yaml.Unmarshal(contents, &c); err != nil {
-		return Config{}, errors.Wrap(err, "unmarshaling config YAML")
+
+	var conf Config
+	err := v.Unmarshal(&conf)
```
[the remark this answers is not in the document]
nirav24: it does.

internal/command/lint/command.go
nirav24: thanks for the suggestion. I'll change it to suggested approach.

internal/config/config.go
```diff
@@ -29,14 +30,26 @@ type Config struct {
 }
 
 // Load loads the config from the given path.
-func Load(path string) (Config, error) {
-	contents, err := ioutil.ReadFile(path)
-	if err != nil {
-		return Config{}, errors.Wrap(err, "reading file")
+func Load(v *viper.Viper) (Config, error) {
```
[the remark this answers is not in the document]
nirav24: apologies, but can you please explain a little bit more?

internal/config/config.go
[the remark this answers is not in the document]
nirav24: I think the reason it does work is I am manually binding it. Let me read up more on mapping it through `mapstructure`

```go
v.BindPFlag("checks.exclude", c.Flags().Lookup("exclude"))
v.BindPFlag("checks.include", c.Flag("include"))
v.BindPFlag("checks.addAllBuiltIn", c.Flag("add-all-built-in"))
v.BindPFlag("checks.doNotAutoAddDefaults", c.Flag("no-defaults"))
```

Thanks for the suggestions

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+			// correctly map config values
+			for i := range subDesc {
+				subDesc[i].Parent = desc.Name
+			}
+
+			if err != nil {
+				return nil, errors.Wrapf(err, "handling field %v", member.Name)
+			}
+			flagDescs = append(flagDescs, subDesc...)
+		case types.Slice:
```
nirav24, line 168: could you please suggest a better way to handle `[]string`?

internal/command/lint/command.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/spf13/viper"
+
```
egurnov, line 8: Looks like this repo uses 2 import groups: standard library and everything else.
Keep these import groups.

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"text/template"
+
+	"github.com/iancoleman/strcase"
+
```
egurnov, line 14: Same, import groups.

internal/config/flags.go
```diff
@@ -0,0 +1,22 @@
...
+func AddFlags(c *cobra.Command, v *viper.Viper) {
+		c.Flags().Bool("add-all-built-in", false, "AddAllBuiltIn, if set, adds all built-in checks. This allows users to explicitly opt-out of checks that are not relevant using Exclude.")
+		v.BindPFlag("checks.add-all-built-in", c.Flags().Lookup("add-all-built-in"))
+		c.Flags().Bool("do-not-auto-add-defaults", false, "DoNotAutoAddDefaults, if set, prevents the automatic addition of default checks.")
+		v.BindPFlag("checks.do-not-auto-add-defaults", c.Flags().Lookup("do-not-auto-add-defaults"))		
+		c.Flags().StringSlice("exclude", nil, "Exclude is a list of check names to exclude.")
+		v.BindPFlag("checks.exclude", c.Flags().Lookup("exclude"))		
+		c.Flags().StringSlice("include", nil, "Include is a list of check names to include. If a check is in both Include and Exclude, Exclude wins.")
+		v.BindPFlag("checks.include", c.Flags().Lookup("include"))
+}
```
egurnov, line 22: How do you make sure this file is regenerated with every change to Config?

I don't see `//go:generate` comment anywhere, so `go-generated-srcs` in Makefile will not pick it up. You probably need to add a target to build your generator first. "Binaries we depend on" section might be the right place.

internal/config/config.go
```diff
@@ -1,42 +1,55 @@
...
-	if err != nil {
-		return Config{}, errors.Wrap(err, "reading file")
+func Load(v *viper.Viper, configPath string) (Config, error) {
+
+	if configPath != "" {
+		filename := filepath.Base(configPath)
+		ext := filepath.Ext(configPath)
+		path := filepath.Dir(configPath)
+
+		v.SetConfigType(strings.TrimPrefix(ext, "."))
```
viswajithiii, line 41: This line is not required, Viper automatically infers it from the extension.

internal/config/flags.go
viswajithiii: Yes, @nirav24, please create a file here similar to https://github.com/stackrox/kube-linter/blob/5e2425e2e1442d1563bad18051c39917c03fa59e/internal/templates/gen.go

egurnov:
There's an e2e test in https://github.com/stackrox/kube-linter/tree/main/e2etests. It would be nice to test this logic in a similar fashion. Although relying on test env having git, internet connectivity and some other repo having linting errors might be a bit unstable. 😕

Having a bad k8s file locally and checking that the same errors are found every time might be a better option.

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+			return errors.Wrapf(err, "couldn't marshal param %v", flagDesc)
+		}
+
+		templateObj = append(templateObj, templateElem{
+			FlagDesc:  flagDesc,
+			ParamJSON: buf.String(),
+		})
+	}
+
+	outFileName := filepath.Join(".", "internal", "config", "flags.go")
```
viswajithiii, line 227: When you add a go generate directive, you will need to make this path relative to the path of the file containing the go generate directive.

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+	Name        string
+	Parent      string
+	Type        flagType
+	Description string
+	AddFlag     bool
+}
+
+type templateElem struct {
+	FlagDesc  flagDesc
+	ParamJSON string
```
viswajithiii, line 92: This ParamJSON seems unused?

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+	"github.com/spf13/viper"
+	"github.com/spf13/cobra"
+)
+
+// AddFlags, walks through config.Check struct and bind its Member to Cobra command 
+// and add respective Viper flag 
+func AddFlags(c *cobra.Command, v *viper.Viper) {
+	{{- range . }}
+		
+	{{- if eq .FlagDesc.AddFlag true }}
```
viswajithiii, line 46:
```suggestion
	{{- if .FlagDesc.AddFlag true }}
```

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+		{{- if eq .FlagDesc.Type "stringSlice" }}		
+		c.Flags().StringSlice("{{ .FlagDesc.Name }}", nil, "{{ .FlagDesc.Description }}")
+		{{- else if eq .FlagDesc.Type "boolean" }}
+		c.Flags().Bool("{{ .FlagDesc.Name }}", false, "{{ .FlagDesc.Description }}")
+		{{- end }}
```
viswajithiii, line 51: Why about the other types? If we don't support them right now that's fine, but we should throw an error so that this doesn't fail silently

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+	{{- end }}
+
+	{{- end }}
+}
+`
+)
+
+var (
+	fileTemplate = template.Must(template.New("gen").Funcs(sprig.TxtFuncMap()).Funcs(template.FuncMap{
+		"backtick": func() string {
```
viswajithiii, line 65: The backtick doesn't seem required here?

internal/config/main/parse.go
viswajithiii: We can just make sure that the underlying type is a string, and return an error if it's not. That way, if someone adds a different kind of parameter, they know they will need to update this code.

internal/config/main/parse.go
```diff
@@ -0,0 +1,244 @@
...
+		return err
+	}
+
+	pkgNames := b.FindPackages()
+	if len(pkgNames) != 1 {
+		return errors.Errorf("found unexpected number of packages in %+v: %d", pkgNames, len(pkgNames))
+	}
+
+	pkg := typeUniverse.Package(pkgNames[0])
+	paramsType := pkg.Type(paramsStructName)
```
viswajithiii, line 204: General thought: I wonder if we can refactor so that we extract out some of this code that's shared with the other codegen code into a common function

viswajithiii (commented):
Thanks for doing this! I've left some comments

viswajithiii:
> There's an e2e test in https://github.com/stackrox/kube-linter/tree/main/e2etests. It would be nice to test this logic in a similar fashion. Although relying on test env having git, internet connectivity and some other repo having linting errors might be a bit unstable. 😕
> 
> Having a bad k8s file locally and checking that the same errors are found every time might be a better option.

Yup, totally agreed. That test is just a sanity test to make sure that the program never crashes. We need to have better end-to-end tests where we actually look at the output and validate things.

nirav24:
@viswajithiii  @egurnov thanks a lot for review. I will make necessary changes.

internal/config/main/parse.go
nirav24: that's a good idea. However, I'd prefer to do that in a separate PR. I will give it a try later.

viswajithiii:
@nirav24 Is this ready for re-review?

nirav24:
> @nirav24 Is this ready for re-review?
> 
> PS: happy new year!

yes, please review it @viswajithiii. Happy new year. 🎊

viswajithiii (approved):
This is great, Nirav! Thanks so much for doing this.

PR #111, opened 2020-12-17
Add unit tests framework

kreamkorokke:
Things in this PR:
1: Changed LintContext from struct to interface, for implementing MockLintContext
2: Added MockLintContext which allows tests to customize the k8s objects that the tests depend on
3: Added some sample test cases to `verify-container-capabilities` check.
4: Fixed a bug within the above check (previous for matching reserved word "all" we did not specify ^ or $)


internal/lintcontext/context.go
```diff
@@ -23,22 +23,37 @@ type InvalidObject struct {
 }
 
 // A LintContext represents the context for a lint run.
-type LintContext struct {
+type LintContext interface {
+	GetObjects() []Object
+	GetInvalidObjects() []InvalidObject
```
viswajithiii, line 28: We should keep the name `Objects()` and `InvalidObjects()`, in keeping with Go convention (https://golang.org/doc/effective_go.html#Getters)

internal/lintcontext/context.go
```diff
@@ -23,22 +23,37 @@ type InvalidObject struct {
...
 	invalidObjects []InvalidObject
 }
 
 // Objects returns the (valid) objects loaded from this LintContext.
-func (l *LintContext) Objects() []Object {
+func (l *lintContextImpl) GetObjects() []Object {
 	return l.objects
 }
 
+// AddObject adds a valid object to this LintContext
```
viswajithiii, line 41: Nit: function name comments should start with with the function name (matching case and all)

internal/lintcontext/context.go
```diff
@@ -23,22 +23,37 @@ type InvalidObject struct {
...
 
+// AddInvalidObject adds an invalid object to this LintContext
+func (l *lintContextImpl) addInvalidObjects(objs ...InvalidObject) {
+	l.invalidObjects = append(l.invalidObjects, objs...)
+}
+
 // New returns a ready-to-use, empty, lint context.
-func New() *LintContext {
-	return &LintContext{}
+func New() *lintContextImpl {
```
viswajithiii, line 57:
```suggestion
func New() LintContext {
```

internal/templates/templates_testutils.go
```diff
@@ -0,0 +1,71 @@
...
+		return actual[i].Message < actual[j].Message
+	})
+
+	for i := 0; i < len(expected); i++ {
+		if expected[i].Message != actual[i].Message {
+			return false
+		}
+	}
+
+	return true
```
viswajithiii, line 70: You can just extract the messages from both and just use `s.ElementsMatch()` on the resulting list, this will create more helpful error messages.

internal/templates/templates_testutils.go
```diff
@@ -0,0 +1,71 @@
...
+	suite.Suite
+
+	Template check.Template
+}
+
+// TestCase represents a single test case which can be verified under a LintContext
+type TestCase struct {
+	Param       interface{}
+	Diagnostics []diagnostic.Diagnostic
+	ExpectError bool
```
viswajithiii, line 24: Name this `ExpectInstantiationError` or so, to make it clear this is an error during instantiation?

viswajithiii (approved):
I like it! All comments are nits

internal/lintcontext/context.go
kreamkorokke: So I actually tried making it LintContext, but then `CreateContexts` also access functions like `loadObjectsFromReader` which is only available to lintContectImpl. I made sure the return values from `CreateContexts` are `[]LintContext`, but wasn't sure if there is another route that I can take here.

viswajithiii: Ahh, that makes sense. The right answer in this case is to unexport the `New` function. (Initially, it was used by outside callers, but after teh `CreateContexts` refactor it doesn't seen required any more).

kreamkorokke: ah yep yep that is true. should just unexport it. will do

PR #112, opened 2020-12-22
Allow users to specify the topology key for the anti-affinity check

viswajithiii:
This allows users to allow inter-pod anti-affinity to use an arbitrary label (example: `topologyKey: failure-domain.beta.kubernetes.io/zone`), not just `kubernetes.io/hostname`.

### DIFF
--- a/internal/templates/templates_testutils.go
+++ b/internal/templates/templates_testutils.go
@@ -1,6 +1,8 @@
 package templates
 
 import (
+	"fmt"
+
 	"github.com/stretchr/testify/suite"
 	"golang.stackrox.io/kube-linter/internal/check"
 	"golang.stackrox.io/kube-linter/internal/diagnostic"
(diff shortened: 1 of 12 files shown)


kreamkorokke (approved):
Now, thought about adding a unit test?;)

kreamkorokke:
And it's fine if you got some other stuffs on hand, we can add them gradually over time as well

internal/templates/templates_testutils.go
```diff
@@ -35,15 +37,18 @@ func (s *TemplateTestSuite) Validate(
...
-			s.Error(err, "param should have caused error but did not raise one")
-			continue
-		}
-		for _, obj := range ctx.Objects() {
-			diagnostics := checkFunc(ctx, obj)
-			s.compareDiagnostics(c.Diagnostics, diagnostics)
-		}
+		s.Run(fmt.Sprintf("%+v", c.Param), func() {
+			checkFunc, err := s.Template.Instantiate(c.Param)
+			if c.ExpectInstantiationError {
```
kreamkorokke, line 42: should this be `c.ExpectInstantiationError && err == nil`?

kreamkorokke (approved):

internal/templates/templates_testutils.go
viswajithiii: `s.Error(err)` inside basically fails if `err != nil` so it's correct.

Issue #114, opened 2020-12-29
Dockerfile: please switch from scratch to something slightly fatter


bootc:
**Description of the problem/feature request**
The Docker image for kube-linter includes nothing but kube-linter itself. This makes for a really small image, but make it unusable in some CI systems such as GitLab CI which require some kind of shell to be present.

**Description of the existing behavior vs. expected behavior**
Without a shell, GitLab CI is unable to use the image to run kube-linter during a CI pipeline. Using a different base image such as the Alpine ones provides enough of a shell to allow it to work.

viswajithiii:
Closing since this is fixed!

bootc:
Thank you!

PR #115, opened 2020-12-29
Dockerfile: switch from scratch to Alpine

bootc:
This adds a shell and other basic system utilities to the image, which
makes kube-linter usable with more CI systems such as GitLab CI that
require a shell. Fixes: #114.

### DIFF
--- a/image/Dockerfile
+++ b/image/Dockerfile
@@ -1,4 +1,4 @@
-FROM scratch
+FROM alpine:latest
 
 COPY ./bin/kube-linter /
 


viswajithiii:
Hi @bootc, thanks for your contribution, and apologies for the slow response as our company was on holiday last week.

I think the use-case for needing a shell makes sense, but I do want to default to the scratch variant since that keeps the image as small as possible, and avoids unnecessary dependencies. So I've decided to take a cue from `nginx` and just add a `-alpine` variant to each of our image tags. Hope that solves your use case.

I wasn't able to push commits to this branch directly, so I've just gone ahead and created #118 off this branch. I'm closing this PR.

PS: happy new year!

PR #118, opened 2021-01-04
Add Alpine base image variant

viswajithiii:
Branched off #115 since I was not able to push commits there directly.

Essentially, create an alpine variant of the Dockerfile and suffix the tag with `-alpine` for those images, similar to what [nginx](https://hub.docker.com/_/nginx) does. Always build and push both images together.


kreamkorokke (approved):
Nice!

PR #121, opened 2021-01-08
Added support for packaged (TGZ) charts

slysunkin:
A feature was added:
Packaged (TGZ) helm chart files are supported

For example, it is possible to run linting like this:
./bin/linux/kube-linter lint mychart-0.1.0.tgz another-test-0.1.0.tgz

### DIFF
--- a/internal/command/lint/command.go
+++ b/internal/command/lint/command.go
@@ -2,10 +2,7 @@ package lint
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
-
-	"encoding/json"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -22,7 +19,6 @@ import (
 func Command() *cobra.Command {
 	var configPath string
 	var verbose bool
-	var jsonPath string
 
 	c := &cobra.Command{
 		Use:   "lint",
@@ -90,22 +86,10 @@ func Command() *cobra.Command {
 					report.FormatPlain(os.Stderr)
 				}
 			}
-			if jsonPath != "" {
-				report := result.GenerateReport()
-				jsonReport, err := json.Marshal(report)
-				if err != nil {
-					return err
-				}
-				err = ioutil.WriteFile(jsonPath, jsonReport, 0644)
-				if err != nil {
-					return err
-				}
-			}
 			return errors.Errorf("found %d lint errors", len(result.Reports))
 		},
 	}
 	c.Flags().StringVar(&configPath, "config", "", "Path to config file")
 	c.Flags().BoolVarP(&verbose, "verbose", "v", false, "Enable verbose logging")
-	c.Flags().StringVar(&jsonPath, "json", "", "Path to output JSON report file")
 	return c
 }
--- a/internal/command/lint/command.go
+++ b/internal/command/lint/command.go
@@ -2,7 +2,10 @@ package lint
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
+
+	"encoding/json"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


internal/lintcontext/parse_yaml.go
```diff
@@ -113,6 +114,79 @@ func (l *lintContextImpl) loadObjectsFromHelmChart(dir string) error {
...
+
+	valuesToRender, err := chartutil.ToRenderValues(chrt, values, chartutil.ReleaseOptions{Name: "test-release", Namespace: "default"}, nil)
+	if err != nil {
+		return nil, err
+	}
+	e := engine.Engine{LintMode: true}
+	rendered, err := e.Render(chrt, valuesToRender)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to render")
+	}
```
viswajithiii, line 175: Can you extract this into a separate function that is reused in `renderHelmChart`?

internal/command/lint/command.go
```diff
@@ -86,10 +90,22 @@ func Command() *cobra.Command {
...
+				if err != nil {
+					return err
+				}
+			}
 			return errors.Errorf("found %d lint errors", len(result.Reports))
 		},
 	}
 	c.Flags().StringVar(&configPath, "config", "", "Path to config file")
 	c.Flags().BoolVarP(&verbose, "verbose", "v", false, "Enable verbose logging")
+	c.Flags().StringVar(&jsonPath, "json", "", "Path to output JSON report file")
```
viswajithiii, line 109: For json output, our plan is to add an output flag instead of taking in a file path. This is already in development, so for simplicity, would you mind removing the JSON report portion of this PR? We're hoping to merge it on our end later this week.

(I will take a look at the code you have right now in `run.go` and make sure our PR offers the same level of usability, so thanks for that!)

viswajithiii (approved):
Thanks for submitting this. This looks good, I just had a couple of minor comments. Please fix and I'll merge it.

internal/lintcontext/parse_yaml.go
slysunkin: Updated

internal/command/lint/command.go
slysunkin: Removed JSON output. I look forward to see the updated output flag.

internal/lintcontext/create_contexts.go
```diff
@@ -36,6 +37,16 @@ func CreateContexts(filesOrDirs ...string) ([]LintContext, error) {
...
 
+		if strings.HasSuffix(strings.ToLower(fileOrDir), ".tgz") {
+			ctx := new()
+			if err := ctx.loadObjectsFromTgzHelmChart(fileOrDir); err != nil {
+				return nil, err
+			}
+
+			contextsByDir[fileOrDir] = ctx
+			continue
+		}
```
viswajithiii, line 48: Ah, sorry for missing this before. This whole block should be moved inside the `filepath.Walk` below, since we want people to just be able to run `kube-linter lint .` and crawl all the subdirectories. As written, this will only load tgz Helm charts if the user explicitly passes the path in the args.

slysunkin: Hi Viswajith, thanks for noticing it. I've updated the code.

internal/lintcontext/create_contexts.go
```diff
@@ -37,21 +37,21 @@ func CreateContexts(filesOrDirs ...string) ([]LintContext, error) {
...
-			continue
-		}
-
 		err := filepath.Walk(fileOrDir, func(currentPath string, info os.FileInfo, walkErr error) error {
 			if walkErr != nil {
 				return walkErr
 			}
 			if !info.IsDir() {
+				if strings.HasSuffix(strings.ToLower(currentPath), ".tgz") {
+					ctx := new()
```
viswajithiii, line 46: Add deduplication to the top (since otherwise, if this same file is encountered through different arguments passed by the user, we would duplicate the work)
```go
if _, exists := contextsByDir[currentPath]; exists {
  return nil
}
```

viswajithiii:
Thanks so much, @slysunkin!

PR #138, opened 2021-03-16
Upgrade to use v0.20.4 dependencies

sbose78:
Not sure how cutting/bleeding edge we'd like to be 🤔


viswajithiii (approved):
Thanks @sbose78, this is helpful!

PR #143, opened 2021-03-26
Create .pre-commit-hooks.yaml

### LINKED ISSUE #142: [FEATURE_REQUEST] Add pre-commit hooks
KubeLinter is a fantastic tool to help you in developing your helm charts and yaml files. It would be good for developers
to have KubeLinter automatically run when committing their code to make sure their charts/yaml follows
best practices. This ensures that the developer does not need to run KubeLinter manually or wait for
CI/CD to run it for them.

The most widely used way of automating the calling of utilities during is the [pre-commit utility](https://pre-commit.com/).
It allows you to easily define, install, and run different utilities using git hooks (the most common being pre-commit hooks).

This requires no great addition to the KubeLinter project, just a simple yaml file at the base of the repo (.pre-commit-hooks.yaml) that defines what the hooks are and how they are called.

By enabling this for the KubeLinter project, it allows developers to automate calling KubeLinter locally in an automated fashion,
decreasing developer feedback time.

jtcarnes:
Added pre-commit hooks for system and golang installs.

Closes #142

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,13 @@
+-   id: kube-linter
+    name: Kube-Linter
+    description: This hook installs (using Go) and runs the Kube-Linter utility to lint Helm charts and Kubernetes Yaml files.
+    entry: kube-linter
+    language: golang
+    types: [yaml]
+  
+-   id: kube-linter-system
+    name: Kube-Linter-System
+    description: his hook runs the Kube-Linter utility that exists already on the system to lint Helm charts and Kubernetes Yaml files.
+    entry: kube-linter
+    language: system
+    types: [yaml]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -13,7 +13,7 @@
   types: [yaml]
 
 - id: kube-linter-docker
-  name: Kube-Linter-Docker
+  name: KubeLinter Docker
   description: This hook runs kube-linter using the project's official docker image
   language: docker_image
   types: [yaml]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,13 +1,20 @@
--   id: kube-linter
-    name: Kube-Linter
-    description: This hook installs (using Go) and runs the Kube-Linter utility to lint Helm charts and Kubernetes Yaml files.
-    entry: kube-linter
-    language: golang
-    types: [yaml]
-  
--   id: kube-linter-system
-    name: Kube-Linter-System
-    description: his hook runs the Kube-Linter utility that exists already on the system to lint Helm charts and Kubernetes Yaml files.
-    entry: kube-linter
-    language: system
-    types: [yaml]
+- id: kube-linter
+  name: Kube-Linter
+  description: This hook installs (using Go) and runs the Kube-Linter utility to lint Helm charts and Kubernetes Yaml files.
+  entry: kube-linter
+  language: golang
+  types: [yaml]
+
+- id: kube-linter-system
+  name: Kube-Linter-System
+  description: This hook runs the Kube-Linter utility that exists already on the system to lint Helm charts and Kubernetes Yaml files.
+  entry: kube-linter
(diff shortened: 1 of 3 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,13 @@
+-   id: kube-linter
+    name: Kube-Linter
+    description: This hook installs (using Go) and runs the Kube-Linter utility to lint Helm charts and Kubernetes Yaml files.
+    entry: kube-linter
+    language: golang
+    types: [yaml]
```
viswajithiii, line 6: I'm a bit confused. How does this hook know what to do?

viswajithiii (commented):
Hi @jtcarnes, thanks for doing this! I must confess I haven't actually used the `precommit` utility before. Can you explain a bit about what exactly a user needs to do to use these?

.pre-commit-hooks.yaml
jtcarnes: There are two different hooks I defined here: one that installs kube-linter using go, the other leverages kube-linter that has already been installed (language: system)


From pre-commit documentation:
```md
The hook repository must contain go source code. It will be installed via go get ./.... pre-commit will create an isolated GOPATH for each hook and the entry should match an executable which will get installed into the GOPATH's bin directory.
```

So for this hook, the pre-commit will install kube-linter using go and once installed invoke it automatically (entry: kube-linter is exactly what it's going to call when pre-commit is run).

The other hook uses system, meaning it simply invokes the command kube-linter and relies on it being installed already on the users computer (say if they didn't want to have a copy per project or already had it installed on their machine through something like homebrew). That is why there are two and how the hook knows what to do.

The hook knows the location of your repo already as it's specified as the source of the hook.

To give a more concrete example, if I wanted to use this pre-commit hook, I would specify a .pre-commit-config.yaml file at the root of my project that would have at minimum something like this:

```yaml
repos:
-   repo: https://github.com/stackrox/kube-linter
    rev: v0.1.6
    hooks:
    -   id: kube-linter
```

This tells pre-commit a few things:
1. the location of the hook it will try and install (your repo).
   - because the language of the hook is golang, as I mentioned from pre-commit's documentation, it will also use this repo location for it's go get command
2. which of the hooks defined there I want to use (this case the kube-linter hook, not the kube-linter-system hook)


With this file at the root of my repository and pre-commit installed on my machine, it will run kube-linter every time right before I commit. This can be disabled if you want to commit something that you know wont pass, but for the most case you want to make sure your code is good before checking it in.

viswajithiii: Got it, this makes a lot of sense. Thank you for the detailed explanation! I think something like #148 is going to be required for usage like this, so thanks for sending that as well.

I'm happy to merge this, but first, can you add a reference to this in our documentation so our users know this is possible? cc: @gaurav-nelson

gaurav-nelson: Thanks @viswajithiii 

@jtcarnes feel free to add a section on the [Using KubeLinter](https://github.com/stackrox/kube-linter/blob/main/docs/using-kubelinter.md) page, or I can add one up tomorrow (probably in the next 24 hours).

jtcarnes: Ok @viswajithiii @gaurav-nelson , I've added my explanation to the Using kubelinter page. Let me know if it needs some edits to better fit.

viswajithiii (approved):
Few minor comments. Thanks for writing the docs, they look fine to me. We can merge once the comments are addressed.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,20 @@
...
+  entry: kube-linter
+  language: system
+  types: [yaml]
+
+- id: kube-linter-docker
+  name: Kube-Linter-Docker
+  description: This hook runs kube-linter using the projects official docker image
+  language: docker_image
+  types: [yaml]
+  entry: stackrox/kube-linter:latest
```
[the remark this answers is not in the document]
viswajithiii: Going to leave this as is for now.

Issue #141, opened 2021-03-26
[FEATURE_REQUEST] Add support for using a non-default clientSchema


aiyengar2:
**Description of the problem/feature request**
A clear and concise description of the problem, or the proposed new feature.

Currently, the scheme used to unmarshall the JSON objects from the manifest is hard-coded to `scheme.Scheme` here:
https://github.com/stackrox/kube-linter/blob/a64df58c92e9dcab418e2cc9f1e796eae70b97fa/pkg/lintcontext/parse_yaml.go#L33

This scheme is used to deserialize the k8s manifests, so any valid k8s objects that are not recognized by the default scheme (e.g. CustomResources) are unable to be deserialized. 

Instead, would it be possible to modify the function signature for `CreateContexts` to optionally take in a `*runtime.Scheme`?

https://github.com/stackrox/kube-linter/blob/a64df58c92e9dcab418e2cc9f1e796eae70b97fa/pkg/lintcontext/create_contexts.go#L23

From a CLI context, this would probably be set to `nil` or `scheme.Scheme`, but this would be helpful for other packages consuming this code.

**Additional context**

Probably a followup for https://github.com/stackrox/kube-linter/issues/119.

I've been trying to consume kube-linter to create a Helm chart testing framework and I hope to upstream it at some point. Making a one-line change to the code linked above seems to be the only blocker, but if this issue is resolved I can go back to consuming upstream directly instead of forking.

viswajithiii:
@aiyengar2 This makes sense to me. Would you be able to send a PR?

aiyengar2:
@viswajithiii done! https://github.com/stackrox/kube-linter/pull/144

viswajithiii:
Closing since this is done.

PR #144, opened 2021-03-26
Add CreateContextWithOptions

aiyengar2:
This will allow consumers of this package to provide non-default schemes to deserialize from or other advanced options via the options.

Related Issue: https://github.com/stackrox/kube-linter/issues/141

### DIFF
--- a/pkg/lintcontext/create_contexts.go
+++ b/pkg/lintcontext/create_contexts.go
@@ -41,8 +41,7 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 			if _, alreadyExists := contextsByDir["-"]; alreadyExists {
 				continue
 			}
-			ctx := new()
-			ctx.customDecoder = options.CustomDecoder
+			ctx := newCtx(options)
 			if err := ctx.loadObjectsFromReader("<standard input>", os.Stdin); err != nil {
 				return nil, err
 			}
@@ -61,8 +60,7 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 
 			if !info.IsDir() {
 				if strings.HasSuffix(strings.ToLower(currentPath), ".tgz") {
-					ctx := new()
-					ctx.customDecoder = options.CustomDecoder
+					ctx := newCtx(options)
 					if err := ctx.loadObjectsFromTgzHelmChart(currentPath); err != nil {
 						return err
 					}
@@ -76,8 +74,7 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 				if knownYAMLExtensions.Contains(strings.ToLower(filepath.Ext(currentPath))) || fileOrDir == currentPath {
 					ctx := contextsByDir[dirName]
 					if ctx == nil {
-						ctx = new()
-						ctx.customDecoder = options.CustomDecoder
+						ctx = newCtx(options)
 						contextsByDir[dirName] = ctx
 					}
 					if err := ctx.loadObjectsFromYAMLFile(currentPath, info); err != nil {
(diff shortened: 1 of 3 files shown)


go.sum
```diff
@@ -186,7 +186,6 @@ github.com/docker/go-metrics v0.0.0-20180209012529-399ea8c73916/go.mod h1:/u0gXw
 github.com/docker/go-units v0.3.3/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/docker/go-units v0.4.0/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/docker/libtrust v0.0.0-20150114040149-fa567046d9b1/go.mod h1:cyGadeNEkKy96OOhEzfZl+yxihPEzKnqJwvfuSUqbZE=
-github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96 h1:cenwrSVm+Z7QLSV/BsnenAOcDXdX4cMv4wP0B/5QbPg=
```
aiyengar2, line 189: These came up when I ran `go mod tidy`. Can remove it from the PR if you'd prefer not to make these changes.

pkg/lintcontext/scheme.go
```diff
@@ -0,0 +1,13 @@
+package lintcontext
+
+import (
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+)
+
+// UseCustomScheme allows a user to set a scheme that is used for deserializing the k8s objects
+// This can be used to add support for deserializing non-native k8s objects that implement the v1.Object interfeace (e.g. custom resources)
+func UseCustomScheme(s *runtime.Scheme) {
```
viswajithiii, line 10: Hmm, instead of this, I would prefer to make this an argument to `CreateContexts` as you had suggested in your issue. I am hesitant to expose a mechanism for mutating global variables, since that will mean that different parts of a program can have unexpected side-effects on each other.

My suggestion would be as follows:
1. Add a new function called `CreateContextsWithOptions`, that takes in a first parameter which is a struct of options. Currently, the only member of the struct will be a `customDecoder`. This `customDecoder` can be passed through into the `lintContextImpl`, all the way into `parseObjects` (and if the `customDecoder` is `nil`, that falls back to the original decoder).
1. Change the implementation of `CreateContexts` to just call `CreateContextsWithOptions` with an empty `Options{}` struct to avoid code duplication.

This will a) ensure no breakage of backward compatibility for any clients using `CreateContexts` and b) ensure that any further options we may want to add to `CreateContextsWithOptions` require no function signature change, only the addition of fields to the `Options` struct.

Does this make sense?

aiyengar2: yup! Updated PR accordingly

aiyengar2:
@viswajithiii I've followed up with https://github.com/stackrox/kube-linter/pull/146, which exposes an additional option for providing additional values.yamls.

pkg/lintcontext/create_contexts.go
```diff
@@ -30,6 +42,7 @@ func CreateContexts(filesOrDirs ...string) ([]LintContext, error) {
 				continue
 			}
 			ctx := new()
+			ctx.customDecoder = options.CustomDecoder
```
viswajithiii, line 45: Can you pass `Options` into the `new` function and populate it there instead?

Also, this is not your code, but we should rename the function to `newCtx()` or something, since as of now, this will prevent the usage of the builtin `new` function in Go in this package. Can you do that, while we're here?

viswajithiii (approved):
This looks good. Please make the one change and I will merge this in. Thanks for the PR!

pkg/lintcontext/create_contexts.go
aiyengar2: Yup sounds good. Will do!

aiyengar2:
@viswajithiii left it as two commits, feel free to squash and merge it.

Issue #145, opened 2021-03-27
[FEATURE_REQUEST] Add support for using non-default values.yaml


aiyengar2:
Followup to https://github.com/stackrox/kube-linter/issues/141. Once https://github.com/stackrox/kube-linter/pull/144 is merged, we can provide more options for the linter to be able to lint non-default values.yamls.

Issue #49, opened 2020-10-30
Support config files with well-known file names


viswajithiii:
Right now, the config file is passed using the `--config` flag to `kube-linter lint`. In practice, however, users may want to have different configs (with different custom checks and different lists of enabled checks) for different apps, which may be in different directories or different Helm charts. Similar to `.eslintrc` files, we could support config files with a well-known filename, say `.kube-linterrc`, and have `kube-linter lint` automatically read them and merge them into the existing config as it walks directories.

bdsoha:
A great addition to this issue would be allowing the search to also be based on an `env` variable:

```console
export KUBE_LINTER_CONFIG=/path/to/the/config
kube-linter lint .
```

PR #146, opened 2021-03-27
Add support for Options.HelmValuesOptions

aiyengar2:
This will allow consumers of this package to provide non-default Helm values.yamls to lint against.

**Note:**: First commit comes from https://github.com/stackrox/kube-linter/pull/144, which will need to be merged in before this is merged. Alternatively, I can add this commit to that PR and drop this instead.

Related Issue: https://github.com/stackrox/kube-linter/issues/145

### DIFF
--- a/pkg/lintcontext/create_contexts.go
+++ b/pkg/lintcontext/create_contexts.go
@@ -10,6 +10,7 @@ import (
 	"github.com/pkg/errors"
 	"golang.stackrox.io/kube-linter/internal/set"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli/values"
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
@@ -36,6 +37,7 @@ func CreateContexts(filesOrDirs ...string) ([]LintContext, error) {
 // CreateContextsWithOptions creates a context with additional Options
 func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintContext, error) {
 	contextsByDir := make(map[string]*lintContextImpl)
+	contextsByChartDir := make(map[string][]LintContext)
 	for _, fileOrDir := range filesOrDirs {
 		// Stdin
 		if fileOrDir == "-" {
@@ -59,14 +61,17 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 				return nil
 			}
 
+			if _, exists := contextsByChartDir[currentPath]; exists {
+				return nil
+			}
+
 			if !info.IsDir() {
 				if strings.HasSuffix(strings.ToLower(currentPath), ".tgz") {
-					ctx := newCtx(options)
-					if err := ctx.loadObjectsFromTgzHelmChart(currentPath); err != nil {
+					lintCtxs, err := CreateHelmContextsWithOptions(HelmOptions{Options: options, FromArchive: true}, currentPath)
+					if err != nil {
 						return err
 					}
-
-					contextsByDir[currentPath] = ctx
+					contextsByChartDir[currentPath] = lintCtxs
 					return nil
 				}
 
@@ -85,15 +90,11 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 				return nil
 			}
 			if isHelm, _ := chartutil.IsChartDir(currentPath); isHelm {
-				// Path has already been loaded, possibly through another argument. Skip.
-				if _, alreadyExists := contextsByDir[currentPath]; alreadyExists {
-					return nil
-				}
-				ctx := newCtx(options)
-				contextsByDir[currentPath] = ctx
-				if err := ctx.loadObjectsFromHelmChart(currentPath); err != nil {
+				lintCtxs, err := CreateHelmContextsWithOptions(HelmOptions{Options: options, FromDir: true}, currentPath)
+				if err != nil {
 					return err
 				}
+				contextsByChartDir[currentPath] = lintCtxs
 				return filepath.SkipDir
 			}
 			return nil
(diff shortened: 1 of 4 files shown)


pkg/lintcontext/parse_yaml.go
```diff
@@ -80,10 +82,9 @@ func (l *lintContextImpl) renderHelmChart(dir string) (map[string]string, error)
 	if err := chrt.Validate(); err != nil {
 		return nil, err
 	}
-	valOpts := &values.Options{ValueFiles: []string{filepath.Join(dir, "values.yaml")}}
```
aiyengar2, line 83: This code isn't necessary since the Helm parser already uses the default `values.yaml` within the chart when it executes the render.

pkg/lintcontext/create_contexts.go
```diff
@@ -9,27 +9,44 @@ import (
...
 
+// Options represent values that can be provided to modify how objects are parsed to create lint contexts
+type Options struct {
+	// CustomDecoder allows users to supply a non-default decoder to parse k8s objects. This can be used
+	// to allow the linter to create contexts for k8s custom resources
+	CustomDecoder runtime.Decoder
+
+	// HelmValuesOptions provide options for additional values.yamls that can be provided to Helm on loading a chart
+	// These will be ignored for contexts that are not Helm-based
+	HelmValuesOptions values.Options
```
viswajithiii, line 28: So, one piece of complexity here is that `CreateContexts` can potentially create multiple Helm contexts by crawling directories, whereas `HelmValuesOptions` would be different for each Helm chart. What are your thoughts here? How are you using this function?

aiyengar2: Ah you are right. I was only using CreateContexts to create a single context, not multiple.

IMO, we should have `lintContextImpl` continue to take in a single values.Options and update the exported `lintcontext.Options` to have `HelmValuesOptions []values.Options`.

Then, on `CreateContexts`, we assert that `len(filesOrDirs) == len(options.HelmValuesOptions)` or return an error.

Does this sound good to you?

viswajithiii: The problem is that `CreateContexts` is recursive, so `len(filesOrDirs) == len(options.HelmValuesOptions)` is not a sufficient condition. Plus, I prefer to avoid parallel arrays anyway.

Long term, in the context of someone running `kube-linter lint`, I would like to move to a model where the user stores these options as code, along with their Helm chart, and KubeLinter reads them and users them. (See #49.) I also want users to be able to specify _multiple_ `values.Options` for a single Helm chart, too, because it would be useful for a user to lint their Helm chart against multiple different settings of values.

For library usage (which is the immediate use-case here), my suggestion is as follows: create a _separate_ function that takes in one directory as input, which _must_ contain a single Helm chart (if not, it returns an error), and then renders it. This function can take in a new `Options` argument, which includes an array of `HelmValuesOptions`, and also embeds the `Options` you created in #144 in it (so that people can use that to pass `customDecoders`). The function then returns a list of `LintContexts`, one for each `HelmValuesOption` passed in (or, if no `HelmValuesOption` was passed in, it uses the default values.yaml and returns exactly one LintContext). We can avoid duplication by calling this function inside the `isHelm` branch of `CreateContexts`. This seems like it will be compatible with the changes we want to make down the road.

Does this make sense to you?

aiyengar2: > The problem is that CreateContexts is recursive, so len(filesOrDirs) == len(options.HelmValuesOptions) is not a sufficient condition. Plus, I prefer to avoid parallel arrays anyway.

I see your point. I didn't realize that CreateContexts was recursive till you just mentioned it 😅 but that makes sense.

> For library usage (which is the immediate use-case here), my suggestion is as follows: create a separate function that takes in one directory as input, which must contain a single Helm chart (if not, it returns an error), and then renders it.

Sounds good to me! This was how I was using it anyways, so this would be perfect for my solution.

aiyengar2: > Long term, in the context of someone running kube-linter lint, I would like to move to a model where the user stores these options as code, along with their Helm chart, and KubeLinter reads them and users them. (See #49.) I also want users to be able to specify multiple values.Options for a single Helm chart, too, because it would be useful for a user to lint their Helm chart against multiple different settings of values.

This is incredibly similar to what I've been working on FYI. If you are interested in taking a peek behind the curtain, here is my code:
https://github.com/aiyengar2/charts-testing/tree/0dade8345626cd045fa9359e9018e20bebe0899a/testing

TLDR of the design:
- You define a [testSuite](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/testing/builder/suite.go#L21), which replaces the Register code you use for registering templates / checks so that they can be dynamically provided by the user
- You then [parse templates](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/tests/setup/parse.go#L40) into the testSuite, where you also have the ability to [set a custom scheme](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/tests/template/tests/main_test.go#L32) if you are using CRDs in your templates
- You then [add a Test (e.g. `testBuilder`)](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/testing/builder/test.go#L13, specifying the templates it applies on (`All|Include|Exclude`), subsets of resources you want to parse within those templates (`On|OnFilter|OnMatcher`), and any function that satisfies the [reflection-based requirements for a DoFunc](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/testing/tests/do.go#L14-L16)
- Finally, you call Run or RunWithContext

https://github.com/aiyengar2/charts-testing/tree/0dade8345626cd045fa9359e9018e20bebe0899a/tests/common contains some examples of DoFuncs that can be written using this framework (only involves writing a struct and then writing the function based on that struct)

https://github.com/aiyengar2/charts-testing/tree/0dade8345626cd045fa9359e9018e20bebe0899a/tests/template/tests contains example code that uses this framework. `main_test.go` and `example_*_test.go` might be interesting to you.

For repository-specific template configurations:
- While parsing in templates, I make sure that it's [only a single Helm chart](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/tests/setup/parse.go#L60)
- I add each Helm chart + values.yamls pair as a separate template [here](https://github.com/aiyengar2/charts-testing/blob/0dade8345626cd045fa9359e9018e20bebe0899a/tests/setup/parse.go#L37)

---

w.r.t. the existing code in kube-linter, you could translate each template generically into a DoFunc based struct + function and convert all your built-in tests into Tests within a single testSuite.

viswajithiii: Ooh this is pretty cool. Definitely looking forward to seeing where this ends up, and learning from it to figure out what makes sense for us to do in this repo (ie, to enable this functionality completely via options to the CLI).

viswajithiii (approved):
Looks good. Can you rebase on `main`? I can merge after that.

viswajithiii:
Hi @aiyengar2, just wanted to check in on this. Will you be able to rebase and merge this one?

nilic:
Any chance for merging this PR?

janisz:
What's the status of this?

janisz:
Closing due to lack of activity

PR #150, opened 2021-03-29
Fix bug where some options in the config file were not honoured

### LINKED ISSUE #149: [BUG] kube-linter v0.1.6 does not respect file based configuration properties.
**System info:**
 - OS: Windows

**Describe the bug**

kube-linter v0.1.6 does not respect file based configuration properties.

**To Reproduce**

Create a yaml config such as the following from the [disabling all default checks example](https://docs.kubelinter.io/#/configuring-kubelinter?id=disable-all-default-checks)

```yaml
checks:
  doNotAutoAddDefaults: true
```

**Expected behavior**

```bash
$ kube-linter.exe lint my-charts/ --config kubelinter-config.yaml
Warning: no checks enabled.
```

**Screenshots**

No screenshot available but the deafult checks do run although I expected them not to.

**Additional context**

I do not have this problem with 0.1.4 or 0.1.5 using the exact same config file.  I do see marshalling errors if I pass a bad config in. CLI arguments suchas `--do-not-auto-add-defaults` work as expected.

viswajithiii:
Fixes #149. Due to incorrect binding from Cobra flags to Viper flags, we started looking for `kebab-case` keys in config, as opposed to `camelCase` keys. Fix the offending code (by fixing the code that generates it).


pkg/config/codegen/parse.go
```diff
@@ -0,0 +1,251 @@
+package main
```
viswajithiii, line 1: This file is the main diff (it was renamed from `main/parse.go`, but unfortunately GitHub is not smart enough to tell)

pkg/config/codegen/parse.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return ""
+}
+
+func constructFlagDescsFromStruct(typeSpec *types.Type, jsonPathSoFar []string) ([]flagDesc, error) {
+	var flagDescs []flagDesc
+	for _, member := range types.FlattenMembers(typeSpec.Members) {
+		extractedTags := types.ExtractCommentTags("+", member.CommentLines)
+		flagNameTags := extractedTags[flagNameTagKey]
+		if len(flagNameTags) > 1 {
```
md2119, line 145: Does `types.ExtractCommentTags` handle comma-separated list? If not, and since we are anticipating 1:1 mapping, maybe we should handle that.

viswajithiii: No, it doesn't treat comma separated lists specially. If a user wants to specify multiple values, they need to do it in multiple lines. Does this make sense? What do you mean by handling that in this context?

md2119: Got it.

md2119 (approved):

Issue #25, opened 2020-10-29
Support OpenShift objects


viswajithiii:
Support checks on OpenShift objects. Suggestions we [got](https://www.reddit.com/r/kubernetes/comments/jjpqn5/introducing_kubelinter_an_open_source_linter_for/gafycbk) were to look at `imagestreams`, `deploymentconfigs`, `buildconfigs` and `routes`.

mancubus77:
+1 for the feature

garethahealy:
@viswajithiii ; any plans to support `Template`? in the same way, k8s `List` and `helm` are.

viswajithiii:
> @viswajithiii ; any plans to support `Template`? in the same way, k8s `List` and `helm` are.

Hmm, I'm not familiar with `Template`. Can you elaborate on how it's used?

garethahealy:
They are a way to provide a simple template for a list of resources (_pre-helm, started in OCP3_). Obviously, customers are migrating to better/other ways, but they are still used by a large number of customers.
- https://docs.openshift.com/container-platform/4.8/openshift_images/using-templates.html

```
cat << EOF > template.yaml
apiVersion: template.openshift.io/v1
kind: Template
metadata:
  name: redis-template
  annotations:
    description: "Description"
    iconClass: "icon-redis"
    tags: "database,nosql"
objects:
- apiVersion: v1
  kind: Pod
  metadata:
    name: redis-master
  spec:
    containers:
    - env:
      - name: REDIS_PASSWORD
        value: ${REDIS_PASSWORD}
      image: dockerfile/redis
      name: master
      ports:
      - containerPort: 6379
        protocol: TCP
parameters:
- description: Password used for Redis authentication
  from: '[A-Z0-9]{8}'
  generate: expression
  name: REDIS_PASSWORD
EOF

oc process --local -f template.yaml
```

viswajithiii:
Hmm, got it. Interesting. It's definitely worth tracking, but as low priority -- meaning we are unlikely to do it anytime soon internally, but we will accept a PR if someone sends one our way.

jfroment:
There is a workaround to make OpenShift templates work with kube-linter: just transform the file using basic jq and/or yq commands, which is easily feasible in an automated environment.

An exemple:
```sh
oc process --local -f your-openshift-template.yaml \
  -p NAME="some_name" \
  -p ENV="prod" \
  -p IMAGE="myregistry.mycompany.com/image:tag" \
[...]

PR #153, opened 2021-04-03
Add OpenShift DeploymentConfig schema

mancubus77:
Add OpenShift DeploymentConfig to support linting. Issue #25

### DIFF
--- a/pkg/lintcontext/parse_yaml.go
+++ b/pkg/lintcontext/parse_yaml.go
@@ -34,8 +34,9 @@ const (
 var (
 	clientSchema  = scheme.Scheme
 	SchemeBuilder = runtime.NewSchemeBuilder(ocsAppsV1.AddToScheme)
-	decoder_error = SchemeBuilder.AddToScheme(clientSchema)
-	decoder       = serializer.NewCodecFactory(clientSchema).UniversalDeserializer()
+	// Catch an error during schema build
+	decoderError = SchemeBuilder.AddToScheme(clientSchema)
+	decoder      = serializer.NewCodecFactory(clientSchema).UniversalDeserializer()
 )
 
 func parseObjects(data []byte, d runtime.Decoder) ([]k8sutil.Object, error) {
--- a/pkg/lintcontext/parse_yaml.go
+++ b/pkg/lintcontext/parse_yaml.go
@@ -32,11 +32,11 @@ const (
 )
 
 var (
-	clientSchema  = scheme.Scheme
+	clientSchema = scheme.Scheme
+	// SchemeBuilder to add OpenShift schema
 	SchemeBuilder = runtime.NewSchemeBuilder(ocsAppsV1.AddToScheme)
-	// Catch an error during schema build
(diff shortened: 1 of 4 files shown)


mancubus77:
e2e tests fail because of environment setup 
image

pkg/lintcontext/parse_yaml.go
```diff
@@ -31,10 +33,16 @@ const (
 
 var (
 	clientSchema = scheme.Scheme
-	decoder      = serializer.NewCodecFactory(clientSchema).UniversalDeserializer()
+	// SchemeBuilder to add OpenShift schema
+	SchemeBuilder = runtime.NewSchemeBuilder(ocsAppsV1.AddToScheme)
+	decoderError  = SchemeBuilder.AddToScheme(clientSchema)
```
viswajithiii, line 38: We can just run this in an `init()` function and `panic` in case there's an error. Since this will be deterministic based on the contents of the `kube-linter` repo, we know we will catch any panics in our automated sanity tests

viswajithiii (approved):
Thanks for doing this! One suggestion around moving this to an `init` function; will merge once that is addressed.

pkg/lintcontext/parse_yaml.go
mancubus77: Done

pkg/lintcontext/parse_yaml.go
```diff
@@ -34,6 +37,14 @@ var (
 	decoder      = serializer.NewCodecFactory(clientSchema).UniversalDeserializer()
 )
 
+func init() {
+	// Add OpenShift schema
+	var SchemeBuilder = runtime.NewSchemeBuilder(ocsAppsV1.AddToScheme)
```
viswajithiii, line 42:
```suggestion
	schemeBuilder := runtime.NewSchemeBuilder(ocsAppsV1.AddToScheme)
```

pkg/lintcontext/parse_yaml.go
```diff
@@ -34,6 +37,14 @@ var (
 	decoder      = serializer.NewCodecFactory(clientSchema).UniversalDeserializer()
```
viswajithiii, line 37: I think you may have to move this call into the `init` function; otherwise, the `init` will get executed _after_ this line, and I don't think the update `schema` will be reflected in the `decoder`. Or were you able to confirm that it successfully worked on a `DeploymentConfig` after this change?

viswajithiii (commented):
Couple comments. Also, can you please rebase and fix the conflicts?

pkg/lintcontext/parse_yaml.go
mancubus77: thanks @viswajithiii 
`decoder` defined as global, so it's within `init()` scope. Yes I tested code as well 
image

pkg/lintcontext/parse_yaml.go
mancubus77: done

viswajithiii:
Thanks, @mancubus77!

PR #156, opened 2021-04-15
Make Checks docs page look similar to the Templates page

gaurav-nelson:
- Updated the template to generate the Checks docs page at https://docs.kubelinter.io/#/generated/checks

### DIFF
--- a/docs/generated/checks.md
+++ b/docs/generated/checks.md
@@ -1,26 +1,304 @@
-# KubeLinter checks
+# KubeLinter cheks
 
 KubeLinter includes the following built-in checks:
 
-| Name | Enabled by default | Description | Remediation | Template | Parameters |
-| ---- | ------------------ | ----------- | ----------- | -------- | ---------- |
- | dangling-service | Yes | Alert on services that don't have any matching deployments | Make sure your service's selector correctly matches the labels on one of your deployments. | dangling-service | `{}` |
- | default-service-account | No | Alert on pods that use the default service account | Create a dedicated service account for your pod. See https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/ for more details. | service-account | `{"serviceAccount":"^(\|default)$"}` |
- | deprecated-service-account-field | Yes | Alert on deployments that use the deprecated serviceAccount field | Use the serviceAccountName field instead of the serviceAccount field. | deprecated-service-account-field | `{}` |
- | drop-net-raw-capability | Yes | Alert on containers not dropping NET_RAW capability | NET_RAW grants an application within the container the ability to craft raw packets, use raw sockets, and it also allows an application to bind to any address. Please specify to drop this capability in the containers under containers security contexts. | verify-container-capabilities | `{"forbiddenCapabilities":["NET_RAW"]}` |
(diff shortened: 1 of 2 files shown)


docs/generated/checks.md
```diff
@@ -1,26 +1,304 @@
...
- | writable-host-mount | No | Alert on containers that mount a host path as writable | If you need to access files on the host, mount them as readOnly. | writable-host-mount | `{}` |
+## dangling-service
+
+**Enabled by default**: Yes
+
+**Description**: Alert on services that don't have any matching deployments
+
+**Remediation**: Make sure your service's selector correctly matches the labels on one of your deployments.
+
+**Template**: [dangling-service](generated/templates.md#dangling-service)
```
gaurav-nelson, line 13: @viswajithiii The links to template page doesn't work as expected where template `HumanName` is different from the template name. Can we fix this somehow so that id's match?

pkg/command/checks/command.go
```diff
@@ -36,13 +36,24 @@ func renderPlain(checks []config.Check, out io.Writer) error { //nolint:unparam
 }
 
 const (
-	markDownTemplateStr = `# KubeLinter checks
+	markDownTemplateStr = `# KubeLinter cheks
```
viswajithiii, line 39:
```suggestion
	markDownTemplateStr = `# KubeLinter checks
```

docs/generated/checks.md
viswajithiii: Hmm, it looks like GitHub doesn't offer you control of the anchor link for headings -- it's going to be derived from the name of the heading, which in our case is the human name.

We can solve this programmatically. I'll comment on the code section.

pkg/command/checks/command.go
```diff
@@ -36,13 +36,24 @@ func renderPlain(checks []config.Check, out io.Writer) error { //nolint:unparam
...
+{{ range . -}}
+## {{ .Check.Name}}
+
+**Enabled by default**: {{ if .Default }}Yes{{ else }}No{{ end }}
+
+**Description**: {{.Check.Description}}
+
+**Remediation**: {{.Check.Remediation}}
+
+**Template**: [{{.Check.Template}}](generated/templates.md#{{.Check.Template}})
```
viswajithiii, line 52: Let's add another field to `augmentedCheck` below called `TemplateLink`, and populate it with the link to the template. The link can be derived as follows:

1. Call [`templates.Get`](https://github.com/stackrox/kube-linter/blob/main/pkg/templates/registry.go#L24) with `chk.Template` for each check. This will return the full template.
1.  Do `strings.Join(strings.Fields(strings.ToLower(template.HumanName)), "-")` to derive the TemplateLink
1. Use template link here (should look like the suggestion below)

```suggestion
**Template**: [{{.Check.Template}}](generated/templates.md#{{.TemplateLink}})
```

gaurav-nelson: Thanks @viswajithiii for explanation, I tried to use this but couldn't  fix it. Can you fix this one?

viswajithiii: Sure!

viswajithiii (approved):
LGTM! @gaurav-nelson if this looks good to you as well, we can merge it

PR #131, opened 2021-02-11
Add JSON output to `lint` and other commands

msugakov:
This adds JSON output for internal structs almost 1:1.

My plan is to (try) attack SARIF in a follow-up PR immediately after this one.

PR #160, opened 2021-04-20
Add SARIF output

msugakov:
This follows up on conversation in the previous PR https://github.com/stackrox/kube-linter/pull/131#issuecomment-814299646

Adding [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html) standard format output should be helpful for `kube-linter` when integrating checks in GitHub Actions and other tools that understand the standard.

I added test as GitHub actions check because it is ultimately what we care about. Also, this shows how to use `kube-linter`'s sarif output with GitHub actions.
Unfortunately, the provided sarif upload step `github/codeql-action/upload-sarif@v1` does not seem to do any validations. Therefore I added JSON schema validation with https://github.com/neilpa/yajsv package.

I also tried Microsoft's https://www.nuget.org/packages/Sarif.Multitool/, a CLI capable of validating sarif files. It can indeed validate the file and produce another SARIF output with results. :) Notably it seems to always return no-zero exit code even on broken files.
I did not go for it because JSON schema validation can probably do majority of checks and I was not able to find how to enable GitHub `GHxxxx` rules in Sarif.Multitool to truly benefit from this tool.

Test scan results are available in [Code scanning alerts](https://github.com/stackrox/kube-linter/security/code-scanning) section under Security tab.
Specifically for this branch, the test results are here: https://github.com/stackrox/kube-linter/security/code-scanning?query=ref%3Arefs%2Fheads%2Fmisha%2Fadd-sarif-output

Also, one of GitHub Actions steps dumps the SARIF output file so you can check what's inside there.


pkg/command/lint/sarif_format.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// TODO: have kube-linter and github address somewhere in "more common" place for such strings.
+	sarifRun := sarif.NewRun("kube-linter", "https://github.com/stackrox/kube-linter")
+	sarifReport.AddRun(sarifRun)
+
+	for _, r := range result.Reports {
+		k8sObjectName := r.Object.GetK8sObjectName()
+		sarifLocation := sarif.NewLocation()
+		sarifLocation.LogicalLocations = append(sarifLocation.LogicalLocations,
+			sarif.NewLogicalLocation().WithName(k8sObjectName.Name).WithKind("k8s object name"),
```
viswajithiii, line 34: This looks kind of weird. Just "Object Name" or "Kubernetes Object Name"?

pkg/command/lint/sarif_format.go
```diff
@@ -0,0 +1,56 @@
...
+			sarif.NewLogicalLocation().WithName(k8sObjectName.Namespace).WithKind("k8s namespace"),
```
viswajithiii, line 35: Same.

pkg/command/lint/sarif_format.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+func formatSarif(out io.Writer, result run.Result) error {
+	sarifReport, err := sarif.New(sarif.Version210)
+	if err != nil {
+		return err
+	}
+
+	// TODO: have kube-linter and github address somewhere in "more common" place for such strings.
+	sarifRun := sarif.NewRun("kube-linter", "https://github.com/stackrox/kube-linter")
```
viswajithiii, line 27: Is there a spot to put the `kube-linter` version in the output too?

viswajithiii:
@msugakov Looks good to me, but I'm obviously not the most familiar with SARIF. Is there some tool we can validate this against? For example, some tool that understands SARIF output and does some visualization or something?

docs/generated/checks.md
```diff
@@ -10,7 +10,7 @@ KubeLinter includes the following built-in checks:
 
 **Remediation**: Make sure your service's selector correctly matches the labels on one of your deployments.
 
-**Template**: [dangling-service](generated/templates.md#dangling-services)
+**Template**: [dangling-service](templates.md#dangling-services)
```
viswajithiii, line 13: Hmm, we keep it this way because this is what seems to work on https://docs.kubelinter.io/

internal/consts/consts.go
```diff
@@ -0,0 +1,7 @@
+package consts
+
+const (
+	ProgramName = "kube-linter"
+	MainURL = "https://github.com/stackrox/kube-linter"
+	TemplatesURL = "https://github.com/stackrox/kube-linter/blob/main/docs/generated/templates.md"
```
viswajithiii, line 6: Link to docks.kubelinter.io?

viswajithiii (commented):
Overall, this looks pretty great to me!

docs/generated/checks.md
msugakov: I don't know how docs.kubelinter.io are built but since that's the main spot for docs which I did not know about, I will not fix what's not broken :) Dropped the commit.

pkg/command/lint/sarif_format.go
msugakov: Renamed to "Object Name".

pkg/command/lint/sarif_format.go
msugakov: Added.

internal/consts/consts.go
msugakov: Done.

msugakov:
@viswajithiii please check the updated code. The thing I'm most unsure about is the location of the test input file: `e2etests/testdata/inputs/splunk.yaml`. Maybe you have a suggestion for a better place for it?

.github/workflows/test-sarif.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Check if output file is valid according to SARIF schema
+        run: |
+          set -ex
+          GO111MODULE=on go get github.com/neilpa/yajsv@v1.4.0
+          schema=$(jq -r '.["$schema"]' results.sarif)
+          [ -n "$schema" ] && wget -nv $schema
+          yajsv -s $(basename $schema) results.sarif
+
+      - name: Upload SARIF output file to GitHub
```
viswajithiii, line 58: Is this something we want to do on every commit? Seems like it was useful for your testing, but in steady state I don't think it makes sense?

.github/workflows/test-sarif.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Build binaries
+        run: make build
+
+      - name: Print kube-linter version
+        run: .gobin/kube-linter version
+
+      - name: Run kube-linter on a sample file with SARIF output
+        run: .gobin/kube-linter lint --format=sarif e2etests/testdata/inputs/splunk.yaml > results.sarif
+        continue-on-error: true
```
viswajithiii, line 38: We do a bunch of this stuff already in CircleCI. I wonder if we can just move this test over there, into the `e2e-tests`?

.github/workflows/test-sarif.yaml
```diff
@@ -0,0 +1,61 @@
...
+            ${{ runner.os }}-go-
+
+      - name: Build binaries
+        run: make build
+
+      - name: Print kube-linter version
+        run: .gobin/kube-linter version
+
+      - name: Run kube-linter on a sample file with SARIF output
+        run: .gobin/kube-linter lint --format=sarif e2etests/testdata/inputs/splunk.yaml > results.sarif
```
viswajithiii, line 37: This file location should be reserve for Go tests written in the `e2etests` directory, IMO. I would suggest just creating another directory called `tests` or so in the root of the repo and putting this in a subdirectory under there.

internal/consts/consts.go
```diff
@@ -0,0 +1,10 @@
+package consts
+
+const (
+	// ProgramName is for displaying help, etc.
+	ProgramName = "kube-linter"
+	// MainURL is where the project info can be found.
+	MainURL = "https://github.com/stackrox/kube-linter"
+	// TemplateURL when formatted with template id, allows user to see help for the given template.
+	TemplateURL = "https://docs.kubelinter.io/#/generated/templates?id=%s"
```
viswajithiii, line 9: Call this `TemplateURLFormat` to make it clear it's not a fully rendered string?

pkg/command/common/format_wrapper.go
```diff
@@ -15,8 +15,11 @@ const (
 	PlainFormat = "plain"
 	// MarkdownFormat is for markdown output suitable for `*.md` files.
 	MarkdownFormat = "markdown"
-	// JSONFormat is for JSON output.
+	// JSONFormat is for JSON output with kube-linter own data structs.
```
viswajithiii, line 18:
```suggestion
	// JSONFormat is for JSON output with kube-linter's own data structs.
```

pkg/run/run.go
```diff
@@ -6,44 +6,51 @@ import (
...
 	Reports []diagnostic.WithContext
 	Summary Summary
 }
 
 // Summary holds information about the linter run overall.
 type Summary struct {
-	ChecksStatus      checkStatus
-	CheckTime         time.Time
+	ChecksStatus      CheckStatus
+	CheckEndTime      time.Time
```
viswajithiii, line 36: Hmm this is a backwards-incompatible change, both for library users and JSON output users. I would suggest to just keep it as `CheckTime`.

msugakov: Yes, it is breaking change but JSON output was merged only 9 days ago https://github.com/stackrox/kube-linter/commit/f7e6f6e7941eca677299b3007626664272ffc9f0#diff-bb95634405f8d047841128dbd79e9058b927a35a23011aa7096d899c21764896R31
I hope there are not many users of `CheckTime` specifically.

.github/workflows/test-sarif.yaml
msugakov: The key is the last step which runs `github/codeql-action/upload-sarif@v1` action.

I can't find a "mainstream" way to call a github action from CircleCI so that we could move this workflow in CircleCI entirely. In the end, it is probably just some nodejs code and I can look for way to invoke it stand-alone if you think it's worth it.

`github/codeql-action/upload-sarif@v1` calls some REST endpoint and theoretically we could call it from CircleCI. However it also generates `partialFingerprints` for de-duplicating alerts which logic I don't think is worth replicating by hand.

> If you upload a SARIF file using the upload-sarif action and this data is missing, GitHub attempts to populate the partialFingerprints field from the source files.
> If you upload a SARIF file without fingerprint data using the /code-scanning/sarifs API endpoint, the code scanning alerts will be processed and displayed, but users may see duplicate alerts.

(from https://docs.github.com/en/code-security/secure-coding/sarif-support-for-code-scanning#preventing-duplicate-alerts-using-fingerprints)

A third option could be to run all steps in CircleCI and only the last one, SARIF upload, in GitHub Actions. For that GitHub Actions workflow should wait for CircleCI workflow and download `results.sarif` artifact when it is ready. Could be possible. I did not go this way initially because the entire GitHub actions workflow takes only 2m 40s.

What do you think?

internal/consts/consts.go
msugakov: Renamed.

.github/workflows/test-sarif.yaml
msugakov: Thanks for suggestion. Moved to a new `tests/testdata` directory.

.github/workflows/test-sarif.yaml
viswajithiii: Hmm, okay with me. We can always remove it later if it seems noisy.

.github/workflows/test-sarif.yaml
viswajithiii: Let's leave it as is. My initial suggestion was predicated on the removal of the upload-sarif step. We should probably eventually migrate all CI here off CircleCI and into GitHub Actions, because there's a bunch of useful actions we likely want to use.

pkg/run/run.go
viswajithiii: Hmm, I doubt there are too, but in my experience, you can sometimes be surprised by how users start adopt specific things. In any case, I'm okay with the risk here.

viswajithiii (approved):

PR #162, opened 2021-04-24
Fix regression on Windows; run sanity tests on Windows

### LINKED ISSUE #161: [BUG]: Unable to use kube-linter in windows
**System info:**
 - Windows

**Describe the bug**
C:\XXX\Kube-Linter>kube-linter lint kube-test1.yaml
**Error: UNEXPECTED: failed to load built-in checks: loading file dangling-service.yaml: open yamls\dangling-service.yaml: file does not exist**

**To Reproduce**
Installed kube-linter in windows and it failed with the above issue

viswajithiii:
Fixes #161

### DIFF
--- a/pkg/builtinchecks/built_in_checks.go
+++ b/pkg/builtinchecks/built_in_checks.go
@@ -2,6 +2,7 @@ package builtinchecks
 
 import (
 	"embed"
+	"fmt"
 	"path/filepath"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


RTann (approved):

pkg/builtinchecks/built_in_checks.go
```diff
@@ -47,7 +48,9 @@ func List() ([]config.Check, error) {
 				loadErr = errors.Errorf("found unexpected entry %s in yamls directory", entry.Name())
 				return
 			}
-			contents, err := yamlFiles.ReadFile(filepath.Join("yamls", entry.Name()))
+			// Do NOT use filepath.Join here, because embed always uses `/` as the separator,
+			// irrespective of the OS we're running.
+			contents, err := yamlFiles.ReadFile(fmt.Sprintf("yamls/%s", entry.Name()))
```
misberner, line 53: Too late, but I think it would have been better to simply use `path.Join` here rather than `fmt.Sprintf`

PR #165, opened 2021-04-29
Make sure lint checks appear in GitHub Pull Request UI

msugakov:
Turns out there's a way to see lint results in PR UI, not only in Security tab.
image

This behavior isn't very well documented and wouldn't work out of the box for `push` event type (like it did not originally). In that case users should consider something like this action to find out PR number: https://github.com/marketplace/actions/find-current-pull-request. Next, it should be put in `github.ref`, but that's beyond my today's investigation capacity.

Many thanks to @misberner for being persistent and showing me [an example](https://github.com/LearnLib/alex/pull/101/files) that it is indeed possible to have lint errors in PR view.

### DIFF
--- a/.github/workflows/test-sarif.yaml
+++ b/.github/workflows/test-sarif.yaml
@@ -1,7 +1,16 @@
 name: Test kube-linter SARIF output
 
 on:
-  - push
+  # Important! Event type should be (or include) `pull_request` if you want to see linter checks in pull request UI.
+  #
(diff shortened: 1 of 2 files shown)


viswajithiii (approved):
This is amazing!! 

Can we get this behaviour wired in to https://github.com/stackrox/kube-linter-action? @neilcar for SA

msugakov:
Will check kube-linter-action

Issue #135, opened 2021-02-21
[BUG] Rule default-service-account is triggered even when automountServiceAccountToken is set to false


smuda:
**System info:**
 - OS: MaxOS

**Describe the bug**
The rule default-service-account is triggered even when automountServiceAccountToken is set to false.

**To Reproduce**
1. Create a deployment with the following content:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kube-linter-test
  labels:
    label1: content
spec:
  replicas: 1
  selector:
    matchLabels:
      label1: content
  template:
    metadata:
      labels:
        label1: content
    spec:
      automountServiceAccountToken: false
      containers:
        - name: kube-linter-test-container
          image: "image/image:tag"
```

Run kube-linter:
`kube-linter lint --do-not-auto-add-defaults --include default-service-account  ../kube-linter-test`

**Expected behavior**
Since "automountServiceAccountToken" is set to false, I'd expect to not trigger the rule even though serviceAccount is not set.

It's my understanding that the default service account should not be used and, only if there is need, a dedicated service account should be used. Having no rights at all (=no account) is better for a pod which does not need them.

PR #166, opened 2021-04-30
ADD: skip when AutomountServiceAccountToken=false

nirav24:
fixes https://github.com/stackrox/kube-linter/issues/135

### DIFF
--- a/pkg/templates/serviceaccount/template_test.go
+++ b/pkg/templates/serviceaccount/template_test.go
@@ -1,0 +1,66 @@
+package serviceaccount
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"golang.stackrox.io/kube-linter/internal/pointers"
+	"golang.stackrox.io/kube-linter/pkg/diagnostic"
+	"golang.stackrox.io/kube-linter/pkg/lintcontext/mocks"
+	"golang.stackrox.io/kube-linter/pkg/templates"
+	"golang.stackrox.io/kube-linter/pkg/templates/serviceaccount/internal/params"
+	appsV1 "k8s.io/api/apps/v1"
+)
+
+func TestServiceAccount(t *testing.T) {
+	suite.Run(t, new(ServiceAccountTestSuite))
+}
+
+type ServiceAccountTestSuite struct {
+	templates.TemplateTestSuite
+
+	ctx *mocks.MockLintContext
+}
+
+func (s *ServiceAccountTestSuite) SetupTest() {
+	s.Init(templateKey)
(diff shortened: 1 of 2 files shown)


viswajithiii (commented):
Thanks for doing this!

We're generally trying to improve our unit test coverage. Would you be open to writing a unit test for this? [This file](https://github.com/stackrox/kube-linter/blob/main/pkg/templates/antiaffinity/template_test.go) should have a pattern you can follow. (Unfortunately, this template has no unit tests right now, so you'll have to write some of the boilerplate too. If you'd prefer not to, I'm also happy to merge this in since the change itself is pretty straightforward -- let me know.)

nirav24:
@viswajithiii I was trying to figure out how to write a test. I will look at the referenced file and give it a shot.

nirav24:
@viswajithiii I've added some tests. Would you please check whether that's how you'd like?

pkg/templates/serviceaccount/template_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Param: params.Params{
+				ServiceAccount: "non-default",
+			},
+			Diagnostics: map[string][]diagnostic.Diagnostic{
+				nonDefaultServiceAccount: {
+					{Message: "found matching serviceAccount (\"non-default\")"},
+				},
+				noAutoMountServiceAccountToken: {},
+			},
+			ExpectInstantiationError: false,
```
viswajithiii, line 57: Right now, we are testing:
1. matching service account + autoMount set to true explicitly => violation
2. non-matching service account + autoMount set to false explicitly => no violation

We should also test the other cases in the matrix, ie:
1. matching service account + autoMount set to `nil` => violation
2. matching service account + autoMount set to `false` explicitly => no violation 
3. non-matching service account + autoMount set to `true` explicitly => no violation
4. non-matching service account + autoMount set to `nil` => no violation

pkg/templates/serviceaccount/template_test.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+		{
+			Param: params.Params{
+				ServiceAccount: "[a)", // Wrong Regex which should raise an error
```
viswajithiii, line 61: 👍

pkg/templates/serviceaccount/template_test.go
nirav24: I will add other tests. Thanks for the suggestion. Appreciate it.

viswajithiii (approved):
Perfect. Thanks you so much @nirav24!

PR #168, opened 2021-04-30
DO NOT MERGE: Test GitHub push trigger and alerts diff

msugakov:
Enable `push` trigger so that GitHub diffs alerts between branches to see how GitHub will show alerts diff


msugakov:
Discarding: used this PR for demoing.

PR #173, opened 2021-05-07
Improved support of kube-linter as a library

slysunkin:
When kube-linter is used as a library (not as the command line utility) it would be more convenient to avoid using temporary file and just pass the stream to the linter.
So the linting process could look like this:
```
// ... load enabled checks for linting ...
// ... for each received chart stream from another app layer:
lintCtxs, err := lintcontext.CreateContextsFromHelmArchive(chartName, chartReader)
result, err := run.Run(lintCtxs, checkRegistry, enabledChecks)
// ... log the linting results
```


viswajithiii (approved):
Seems fine to me

PR #176, opened 2021-05-12
Fixed bad link

gaurav-nelson:
Based on https://kube-linter.slack.com/archives/C01ASRW6Y8L/p1620828189029700

> link to supported objectKinds is incorrect. https://github.com/stackrox/kube-linter/tree/main/internal/objectkinds gives a 404.
> we need to update that link. It should be https://github.com/stackrox/kube-linter/tree/main/pkg/objectkinds. (cc: @Gaurav )

### DIFF
--- a/docs/configuring-kubelinter.md
+++ b/docs/configuring-kubelinter.md
@@ -131,7 +131,7 @@ For example,
           - DeploymentLike
   ```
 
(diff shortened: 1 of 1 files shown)


viswajithiii (approved):
Thanks man!

PR #177, opened 2021-05-14
Move stringutils from internal to pkg

rrati:
The stringutils package has a number of really useful utilities for any consumer of kube-linter that would allow the consumer to use the same standards for any checks they write to conform with those included in kube-linter.  Moving this package out of internal allows consumers to use it.


rrati:
Closing since there are plans to move these functions to a library for general consumption.

PR #180, opened 2021-05-21
Add a method to retrieve all the registered kinds

rrati:


pkg/objectkinds/registry.go
```diff
@@ -18,6 +18,16 @@ func registerObjectKind(name string, objectKind Matcher) {
 	allObjectKinds[name] = objectKind
 }
 
+// AllObjectKinds will return all the object kind names that are registered
+func AllObjectKinds() []string {
+	var kinds []string
```
viswajithiii, line 23:
```suggestion
	kinds := make([]string, 0, len(allObjectKinds))
```

viswajithiii (approved):

pkg/objectkinds/registry.go
rrati: I somehow figured you would want that.  I had it before but removed.  I'll add back

rrati: Ah, nm.  You made the change.

Issue #181, opened 2021-06-01
[FEATURE_REQUEST] Add check for minimum replica count


rrati:
**Description of the problem/feature request**
A durable and reliable service should have a minimum number of replicas set that allows a service to tolerate upgrades or pod difficulty without losing service.  The number of replicas could vary by service, but usually a minimal acceptable replica count is 3.

**Description of the existing behavior vs. expected behavior**
Currently there is a check for pod anti-affinity if the replica count >= 2, but that check doesn't check a minimum safe replica count.  Once this check is implemented, the default would be at 3 but the template would support any number of minimum replicas.  Additionally, this check would not alert on an entity if it is not the controller (so alert on a deployment but not a replicaset created by a deployment).

viswajithiii:
This makes a lot of sense, and is something we'd be happy to accept (should be quite straightforward to implement). I don't think we should enable it by default but we could have a built-in check with 3 replicas to nudge people towards that.

> Additionally, this check would not alert on an entity if it is not the controller (so alert on a deployment but not a replicaset created by a deployment)

Note that this is not something that will affect `kube-linter` itself, since we will only scan YAML files that are present (which will only be for the controller, since the other objects will be created live by the api server). But I know you're using this in a live cluster. Since you're using it as a library, you can probably control this by only passing in objects that are controllers (via looking at `ownerReferences`) -- I think this property of alerting only on the controller is desirable for all checks, not just this one.

rrati:
There's actually some helper functions in apimachinery that tell if you the object is a controller or not.  Good point though, it's easy enough to verify that before passing the object to kube-linter.  I have this check implemented already, so will open a PR with the implementation.

PR #185, opened 2021-06-01
Add check for minimum number of replicas. #181

rrati:


pkg/builtinchecks/yamls/minimum-replicas.yaml
```diff
@@ -0,0 +1,10 @@
+name: "required-minimum-replicas"
```
viswajithiii, line 1: This name should have the number three in it, since that's what it checks for. (The recommendation is not a global standard, but one that we are making, so it seems better to be explicit.)

pkg/templates/replicas/template.go
```diff
@@ -0,0 +1,48 @@
...
+	"golang.stackrox.io/kube-linter/pkg/templates/replicas/internal/params"
+)
+
+const (
+	templateKey = "minimum-replicas"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Minimum recommended replicas not met",
```
viswajithiii, line 23: The human name is typically not a sentence
```suggestion
		HumanName:   "Minimum replicas",
```

pkg/templates/replicas/template.go
```diff
@@ -0,0 +1,48 @@
...
+		Key:         templateKey,
+		Description: "Flag applications running fewer than recommended number of replicas",
```
viswajithiii, line 25:
```suggestion
		Description: "Flag applications running fewer than the specified number of replicas",
```

viswajithiii (commented):
Couple minor comments. Happy to merge after that. Thanks for writing unit tests!

pkg/builtinchecks/yamls/minimum-replicas.yaml
rrati: Fixed

pkg/templates/replicas/template.go
rrati: Fixed

pkg/templates/replicas/template.go
rrati: Fixed

viswajithiii (approved):

Issue #184, opened 2021-06-01
[FEATURE_REQUEST] Add check for UpdateStrategy


rrati:
**Description of the problem/feature request**
An updateStrategy of RollingUpdate along with a minimum unavailable helps ensure a service can be upgraded without disrupting service.  Users may have other configurations they determine would allow for safe upgrades of their service.  The desire here is to implement a check that validates an object has an acceptable upgrade strategy.

**Description of the existing behavior vs. expected behavior**
Once implemented the check would fail any object that does not conform to the update strategy.  By default, rollingUpdate and maxUnavailable for 1 (or 3?) seem to make sense.

viswajithiii:
This is great! Definitely would be a helpful addition.

rrati:
I am currently working on this ticket

PR #190, opened 2021-06-15
Added template for update strategy and a basic check #184

rrati:


pkg/builtinchecks/yamls/upgrade-strategy.yaml
```diff
@@ -0,0 +1,10 @@
+name: "upgrade-strategy"
+description: "Indicates when a deployment doesn't meet safe upgrade parameters"
+remediation: >-
+  Correct the update strategy to perform safe upgrades.
```
viswajithiii, line 4: Can we add more colour here? Safe is a very broad term. We should educate users on why the benefits of RollingUpdate, and perhaps insert a link to k8s documentation.

pkg/builtinchecks/yamls/upgrade-strategy.yaml
```diff
@@ -0,0 +1,10 @@
...
+scope:
+  objectKinds:
+    - DeploymentLike
+template: "upgrade-configuration"
+params:
+  StrategyTypeRegex: "^(RollingUpdate|Rolling)$"
```
viswajithiii, line 10:
```suggestion
  strategyTypeRegex: "^(RollingUpdate|Rolling)$"
```

pkg/builtinchecks/yamls/upgrade-strategy.yaml
```diff
@@ -0,0 +1,10 @@
+name: "upgrade-strategy"
```
viswajithiii, line 1: Should we make the name more specific, like "no-rolling-updates" or something. Also I think we should generally use "update" everywhere (names and text) instead of a mix of "upgrade" and "update".

pkg/templates/upgradeconfig/internal/params/params.go
```diff
@@ -0,0 +1,10 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
+	StrategyTypeRegex  string
+	MaxPodsUnavailable string
+	MinPodsUnavailable string
+	MaxSurge           string
+	MinSurge           string
```
viswajithiii, line 9: Can you add comments here? Those get populated as descriptions in the docs. Also, if any are required, you should add a "// +required"

pkg/extract/update_strategy.go
```diff
@@ -0,0 +1,90 @@
...
+import (
+	"reflect"
+
+	"golang.stackrox.io/kube-linter/pkg/k8sutil"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// UpdateStrategy will extract the UpdateStrategy struct from a provided
+// object if it exists
+func UpdateStrategy(obj k8sutil.Object) (interface{}, bool) {
```
viswajithiii, line 12: Instead of depending on the caller to chain these functions in the appropriate way, what do you think of just having a function like:
```go
func UpdateStrategyFields(obj k8sutil.Object) (type string, maxUnavailable, maxSurge *intstr.IntOrString)
```
and having the function just return these? You can return `type` as `""` if not found, and the others as `nil`. The caller code then becomes much simpler, and all the functions you have here can be unexported.

pkg/templates/upgradeconfig/template.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	actualVal, err := intstr.GetValueFromIntOrPercent(actual, 100, false)
+	if err != nil {
+		return false
+	}
+	actualIsPercent := strings.Contains(actual.String(), "%")
+	if len(max) > 0 {
+		maxIntOrString := intstr.Parse(max)
+		maxIsPercent := strings.Contains(maxIntOrString.String(), "%")
+		if actualIsPercent != maxIsPercent {
```
viswajithiii, line 43: Hmm, why?

viswajithiii (commented):
This looks good, and thanks for the thorough testing. Just had some thoughts on code structure

pkg/templates/upgradeconfig/template.go
rrati: So, as I was thinking about how to handle percent vs fixed, I came up with a few options:
1) Grab the replicas  and do a calculation to determine if a fix MaxPodsUnavailable would meet a percent requirement in the check
2) Force the two entities to be the same type

I went with 2 because while 1 would work for an initial check, it doesn't allow for scaling once deployed.  A fixed number of MaxUnavailable may meet a percent requirement with the initial deployment, but then would not meet it if the deployment is scaled after deployment.  So, I decided if the check is using a percent value then a percent should be enforced on the entity being checked to ensure it stays correct even if the replicas are scaled after deployment

pkg/extract/update_strategy.go
rrati: I originally thought about doing that and headed down that route, but then it makes the extract functions very specific.  I didn't put the implementation in, but StatefulSets have an UpdateStrategy field, but only have it filled with a partition field.  Likewise, DeploymentConfigs have a lot more fields than maxUnavailable and maxSurge.  I ended up with this path to make it easier to reuse the functions for later checks if anything came up.

I suppose in the suggestion you make, the path for anything that wants to extract additional information is to create new extraction functions that use some of the non-exported functions.  Is that the path you'd prefer?

pkg/builtinchecks/yamls/upgrade-strategy.yaml
rrati: Yes, I'll change the check name.  I originally had it enforcing more, but there just isn't going to be a sane set of defaults for anything other than to push using RollingUpdate.

pkg/extract/update_strategy.go
viswajithiii: No, my suggestion was a lot simpler. Basically, right now, you're doing all this in the `template`:

[Call the following code snippet 1]
```go
				strategy, found := extract.UpdateStrategy(object.K8sObject)
				strategyType, found := extract.TypeFromUpdateStrategy(strategy)
				rollingUpdate, found := extract.RollingUpdateFromUpdateStrategy(strategy)
				maxUnavailable, found := extract.MaxUnavailableFromRollingUpdate(rollingUpdate)
				maxSurge, found := extract.MaxSurgeFromRollingUpdate(rollingUpdate)
```

This requires quite a bit of work from the caller, and requires the caller to understand the internals of this code. Instead of this, I would propose that we have a function in the `extract` package like:
```go
func UpdateStrategyFields(obj k8sutil.Object) (type string, maxUnavailable, maxSurge *intstr.IntOrString)
```
internally, this function would just be making the series of calls that I listed above (in code snippet 1). The difference now is that the `template` gets all this in one call.

For better readability and extensibility, we can also change my proposed function's signature so that instead of returning a lot of values, we define a custom struct like `UpdateStrategyData` or whatever, which has all these fields, and then fill those. IMO, that will provide a much cleaner interface for the caller.

Does that make sense?

rrati: That is what I understood you to mean, save the custom struct bit.  You've described the first pass I took at this, but in the end I ended up not liking what the caller had to do with a custom struct to handle the variable results/scenarios iirc.  I've since changed logic of things quite a bit so maybe some of those concerns are gone.

I can refactor this way again and see if the structure of things is better now that the template logic is worked out better.

pkg/builtinchecks/yamls/upgrade-strategy.yaml
rrati: Done

pkg/builtinchecks/yamls/upgrade-strategy.yaml
rrati: Done

pkg/extract/update_strategy.go
rrati: Done

pkg/builtinchecks/yamls/upgrade-strategy.yaml
rrati: I added some more detail here.  Unfortunately it doesn't appear there is a single link to the kube documentation for this because the update strategy configuration is slightly different for all objects.

pkg/templates/upgradeconfig/internal/params/params.go
rrati: Done

pkg/templates/updateconfig/internal/params/params.go
```diff
@@ -0,0 +1,33 @@
...
+type Params struct {
+	// A regular expression the defines the type of update
+	// strategy allowed.
+	// +required
+	StrategyTypeRegex string
+
+	// The maximum value that be set in a RollingUpdate
+	// configuration for the MaxUnavailable.  This can be
+	// an integer or a percent.
+	// +optional
```
viswajithiii, line 13: You can remove `+optional`, that's the default and our code actually doesn't look at this value

pkg/templates/updateconfig/template.go
```diff
@@ -0,0 +1,156 @@
...
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			compiledRegex, err := regexp.Compile(p.StrategyTypeRegex)
+			if err != nil {
+				return nil, errors.Wrapf(err, "invalid regex %s", p.StrategyTypeRegex)
+			}
```
viswajithiii, line 105: It may make sense to parse the `IntOrString`s from the params here too, so we can return an error at check instantiation time if any are wrong? Right now if it's defined incorrectly, we will just silently not fail any checks, which doesn't seem ideal.

viswajithiii (commented):
Couple minor comments but this looks great!

pkg/templates/updateconfig/template.go
rrati: Added checks and tests for configuration

pkg/templates/updateconfig/internal/params/params.go
rrati: Done

pkg/templates/updateconfig/template.go
```diff
@@ -0,0 +1,208 @@
...
+			if len(p.MinSurge) > 0 {
+				minSurge, err = parseIntOrString(p.MinSurge)
+				if err != nil {
+					errorList.AddWrapf(err, "invalid MinSurge %s", p.MinSurge)
+				}
+			}
+			if errorList.ToError() != nil {
+				return nil, errorList.ToError()
+			}
```
viswajithiii, line 157:
```suggestion
			if err := errorList.ToError(); err != nil {
				return nil, err
			}
```

viswajithiii (approved):
Thanks for working through everything!

Issue #183, opened 2021-06-01
[FEATURE_REQUEST] Add check for trusted tags


rrati:
**Description of the problem/feature request**
An image tag may become untrustworthy if it is found to have a severe bug/vulnerability or has been discovered to have been built from compromised sources.  Alternately a user may want to only allow explicit tags to be deployed in their workloads, either to disallow something like the latest tag, or to specify a list of approved tags for an image.

**Description of the existing behavior vs. expected behavior**
If enabled, this check would fail any manifests that use tags that are not configured as acceptable by the check.  This could be a whitelist or a blacklist.

viswajithiii:
Makes sense. This is very similar to #182. I wonder if we should have a template that takes in a list of image names, allowing the user to specify a regex so they can specify a registry, or a tag, or a combination of both. The template can take two lists: an `allowlist` and a `blocklist`, and we can say something like:
1. If `allowlist` is specified, the image must match at least one regex from the allowlist.
2. If `blocklist` is specified, the image must not match any regex from the blocklist.

(This implies that the `blocklist` takes precedence -- if an image is in both the `allowlist` and `blocklist`, then the check will still be violated. That seems like a better stance to take, to err on the side of more caution.)

Let me know what you think. Such a template would solve both #182 and this, and we could then just implement checks for these.

rrati:
I was thinking this was very similar to #182 as I was writing it.  A regex may actually work well here with a single template that will do the work.

I do wonder if we should be more exacting in the template though.  When we allow allow and block lists together, things can get messy.  What the template only allowing one or the other to be specified, and if the user wants an allow and a block list then they define to different checks using the same template?  Then there's no magic to figure out which takes precidence.

viswajithiii:
> I was thinking this was very similar to #182 as I was writing it. A regex may actually work well here with a single template that will do the work.
> 
> I do wonder if we should be more exacting in the template though. When we allow allow and block lists together, things can get messy. What the template only allowing one or the other to be specified, and if the user wants an allow and a block list then they define to different checks using the same template? Then there's no magic to figure out which takes precidence.

Makes sense. From thinking through it, there is no flexibility lost by having them in separate templates.

Priyankasaggu11929:
@rrati, @viswajithiii, thanks for creating the issue. I'm picking this up & will be working on the implementation. Thank you!

janisz:
It was closed by #191

PR #191, opened 2021-06-23
add check for validating improper container image tags

Priyankasaggu11929:
**The PR makes the following changes:**

- adds a new builtin default check `improper-container-image-tag`, for validating `DeploymentLike` kubernetes objects running containers with improper image tags.
- Currently, the check is limited to checking container image with a `latest` floating tag only. But the regex for checking improper tags can be extended further as required.

**Fix issue** https://github.com/stackrox/kube-linter/issues/183

---

The output looks like:

Screenshot from 2021-06-29 08-33-19

Signed-off-by: Priyanka Saggu <priyankasaggu11929@gmail.com>


docs/generated/checks.md
```diff
@@ -2,6 +2,22 @@
 
 KubeLinter includes the following built-in checks:
 
+## check-untrusted-images
```
viswajithiii, line 5: This doesn't seem like a great built-in check. How about just having a check for `latest` tag alone? That seems more focused and useful, and we can enable it by default too.

docs/generated/templates.md
```diff
@@ -613,6 +613,39 @@ KubeLinter supports the following templates:
...
+    "name": "allowedList",
+    "type": "array",
+    "description": "list of allowed container image registries",
+    "required": false,
+    "regexAllowed": true,
+    "negationAllowed": true,
+    "arrayElemType": "string"
+  },
+  {
+    "name": "blockedList",
```
viswajithiii, line 638: Nit: it seems more in line with industry practice to call this `allowlist` and `blocklist`. (Since it's one word, we should lower-case the `L` in `list`)

pkg/templates/untrustedimages/template.go
```diff
@@ -0,0 +1,55 @@
...
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
+
+				// checking whether the container image is a trusted or not
+
```
viswajithiii, line 35: I would just remove the comment and these blank lines, I don't think the comment adds any extra information.
```suggestion
```

pkg/templates/untrustedimages/template.go
```diff
@@ -0,0 +1,55 @@
...
+				allowedResult, blockedResult := false, true
+
+				for _, allowed := range p.AllowedList {
+					allowedRe := regexp.MustCompile(allowed)
```
viswajithiii, line 39: Don't use `MustCompile` here, as that will make `kube-linter` crash at runtime if someone uses a bad regex. Instead, we should compile the regexes at the top of this function, and return an instantiation error. For example, see the pattern [here](https://github.com/stackrox/kube-linter/commit/72b90689e78007030b8aa18513db28a5ec455761#diff-f5177d5dde82cfc5a418885d5ece6774652f07cf0e7ff174d41d06eee1ae169dR41).

pkg/templates/untrustedimages/template.go
```diff
@@ -0,0 +1,55 @@
...
+					allowedResult = allowedRe.MatchString(container.Image)
```
viswajithiii, line 40: This seems wrong. You would want to `||` this with the current `allowedResult`, right?

pkg/templates/untrustedimages/template.go
```diff
@@ -0,0 +1,55 @@
...
+				}
+
+				for _, blocked := range p.BlockedList {
+					blockedRe := regexp.MustCompile(blocked)
+					blockedResult = blockedRe.MatchString(container.Image)
+				}
+
+				if blockedResult || !allowedResult {
+					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("The container %q is using an untrusted image, %q.", container.Name, container.Image)}}
```
viswajithiii, line 49: We should make this message more useful.
For example:
1. If the image is explicitly blocked, then we should say which blocklist regex it matched.
2. If we're failing because it's not explicitly allowed, then we should say that.

pkg/templates/untrustedimages/template.go
```diff
@@ -0,0 +1,55 @@
+package untrustedimages
```
viswajithiii, line 1: Please add unit tests.

docs/generated/templates.md
Priyankasaggu11929: noted. Thank you!

pkg/templates/untrustedimages/template.go
Priyankasaggu11929: removed the comment & lines.

pkg/templates/untrustedimages/template.go
Priyankasaggu11929: I'll be adding unit tests. Thanks!

docs/generated/checks.md
Priyankasaggu11929: Re: `How about just having a check for latest tag alone?`

Would it be useful to break this check into 2 different checks then?

- Keep the first to just check for `latest` tag alone & then add it as a default built-in check.
- And then make a second one that checks for the container image source registry?

Also, do you have pointers or any suggestion to improve the image source registry check part?

Thank you!

pkg/templates/untrustedimages/template.go
Priyankasaggu11929: Thanks for pointing. I'll move it to top & add proper error handling.

pkg/templates/untrustedimages/template.go
Priyankasaggu11929: Noted. Thank you!

pkg/templates/untrustedimages/template.go
Priyankasaggu11929: Sorry, I'm unable to understand this part.

Are you pointing [at this block](https://github.com/stackrox/kube-linter/blob/ddb195fe7855bb7c1482e5d94c2771accdd3ec66/pkg/templates/untrustedimages/template.go#L48-L50)?

```
if blockedResult || !allowedResult {...}
``` 
And could you please explain what is wrong. I'll try to correct it accordingly. Thank you!

docs/generated/checks.md
Priyankasaggu11929: @viswajithiii,, I've implemented the current PR for the `latest` tag check alone. I've added this check as a `built-in default` check, as you suggested above.

Also, have added a `template_test.go` file with required unit tests.

---

Re-iterating: `do you have pointers or any suggestion to improve the image source registry check part?`

Thank you!

pkg/builtinchecks/yamls/improper-container-image-tag.yaml
```diff
@@ -0,0 +1,9 @@
+name: "improper-container-image-tag"
```
viswajithiii, line 1: We should just call this `latest-tag`, yeah?

pkg/templates/impropercontainerimagetag/internal/params/params.go
```diff
@@ -0,0 +1,8 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
+
+	// list of regular expressions for blocked or bad container image tags
+	Blockedlist []string
```
viswajithiii, line 7:
```suggestion
	BlockList []string
```

pkg/templates/impropercontainerimagetag/template.go
```diff
@@ -0,0 +1,74 @@
...
+					return nil
+				}
+
+				var results []diagnostic.Diagnostic
+
+				for _, container := range podSpec.Containers {
+					blockedResult := isInList(blockedRegexes, container.Image)
+					if blockedResult {
+						results = append(results, diagnostic.Diagnostic{Message: fmt.Sprintf("The container %q is using an improper image tag, %q.", container.Name, container.Image)})
+					}
```
viswajithiii, line 55:
```suggestion
					if isInList(blockedRegexes, container.Image) {
						results = append(results, diagnostic.Diagnostic{Message: fmt.Sprintf("The container %q is using an improper image tag, %q.", container.Name, container.Image)})
					}
```

viswajithiii (commented):
Why did you remove the allowlist part?

pkg/builtinchecks/yamls/improper-container-image-tag.yaml
Priyankasaggu11929: Thank you. Yes, I renamed it to "latest-tag"

pkg/templates/impropercontainerimagetag/internal/params/params.go
Priyankasaggu11929: Done!

pkg/templates/impropercontainerimagetag/template.go
Priyankasaggu11929: Done. Thank you!

Priyankasaggu11929:
@viswajithiii , Re: 
> Why did you remove the allowlist part?

Ah, I feel I misunderstood your suggestion from this comment https://github.com/stackrox/kube-linter/pull/191#discussion_r658209334 

This part -> `How about just having a check for latest tag alone?`

I'm thinking of writing a separate check for the allowed image registry sources & will use the 
`AllowList: [" ^(docker.io|quay.io)" ]` param in there, for the same.

Would that be fine in your opinion? Or there could be additions, modifications to improve? Thank you!

Priyankasaggu11929:
Hello @viswajithiii, just a quick follow-up, to check if there are more improvements required on the PR.

Thanks a lot! :)

pkg/builtinchecks/yamls/latest-tag.yaml
```diff
@@ -0,0 +1,9 @@
+name: "latest-tag"
+description: "Indicates when a deployment-like object is running a container with a floating image tag, \"latest\""
+remediation: "Use a container image with a proper image tag, outside the set blocked tag regex \".*:(latest)$\"."
+scope:
+  objectKinds:
+    - DeploymentLike
+template: "latest-tag"
+params:
+  BlockList: [".*:(latest)$" ]
```
viswajithiii, line 9:
```suggestion
  blockList: [".*:(latest)$" ]
```

viswajithiii (approved):
LGTM!

Issue #194, opened 2021-07-12
[FEATURE_REQUEST] adding new checks to make sure a network policy captures pod/s


shireenf-ibm:
adding checks to make sure:

- [x] each NetworkPolicy captures at least one pod 

- [x]  each PodSelector in a NetworkPolicy rule captures at least one pod

Issue #36, opened 2020-10-30
Add checks for schema validation


viswajithiii:
Add checks that ensure to validate objects against the schema. This can be done by checking against the K8s OpenAPI spec. Some preliminary code is in https://github.com/stackrox/kube-linter/tree/viswa/checkpoint-swagger.

sbose78:
This one's going to be helpful!

PR #198, opened 2021-08-05
adding checks that indicates when networkpolicy or networkpolicy rule do not have any asso…

shireenf-ibm:
#194 …ciated deployments... based on their podSelectors


shireenf-ibm:
Hi,
I have added a check that indicates when a Networkpolicy (k8s.object) do not have any associated pods (deploymentLike objects).
It works well (as expected) when the networkpolicy is written correctly (with no typo/syntax/type errors) for all cases.
 so for example, in the following cases it behaves as expected:
- when the networkpolicy's spec podSelector is empty and there are associated pods (podSelector: { } )  - (matches everything - no lint errors)
- when the podSelector has labels and there is a match in one/more pods - (no lint errors)
- when the podSelector has labels and there is no match in any of the associated pods (flags a lint error ).
------------------------------------------------
However, I didn't find out if kube-linter detect 'syntax/type/validation' errors.
 for example :
 - for k8s , a podSelector is a required field and may not be null:
 (for example in live k8s cluster, a missing podSelector will output this error: 

> error validating data: ValidationError(NetworkPolicy.spec): missing required field "podSelector" in io.k8s.api.networking.v1.NetworkPolicySpec;

 )
 _but I couldn't find how to check field existence in kube-Linter_ (missing or null podSelector behaves as if it was empty) 
 - also an invalid podSelector _with wrong keys or wrong type_ didn't output any errors on kube-Linter (I didn't find if there is a way to check these errors too)
 **would like to get your help/suggestions in these cases.**

viswajithiii:
Hi Shireen, my apologies for the extremely late response here. So, couple of things. The functionality you're adding in this PR makes sense and is valuable.

For the case of required fields, that's a known gap, and it's something we will tackle as part of #36.

viswajithiii (approved):
Thanks!

PR #199, opened 2021-08-09
add allowlist parameter to the exisiting latestTag check

Priyankasaggu11929:
The PR does the following:

- Add another parameter `AllowList` to the check with a default value equals to empty list 
- Improve the existing check logic to validate that the check has either `AllowList` or `BlockList` set or both set and then action properly.

### DIFF
--- a/pkg/builtinchecks/yamls/latest-tag.yaml
+++ b/pkg/builtinchecks/yamls/latest-tag.yaml
@@ -1,9 +1,9 @@
 name: "latest-tag"
-description: "Indicates when a deployment-like object is running a container with a floating image tag, \"latest\""
-remediation: "Use a container image with a proper image tag, outside the set blocked tag regex \".*:(latest)$\"."
+description: "Indicates when a deployment-like object is running a container with an invalid container image"
+remediation: "Use a container image with a proper image tag satisfying either the \"AllowList\" & \"BlockList\" regex patterns."
 scope:
   objectKinds:
     - DeploymentLike
 template: "latest-tag"
 params:
-  BlockList: [".*:(latest)$" ]
+  BlockList: [".*:(latest)$"]
--- a/pkg/templates/latesttag/internal/params/params.go
+++ b/pkg/templates/latesttag/internal/params/params.go
@@ -3,6 +3,9 @@ package params
 // Params represents the params accepted by this template.
 type Params struct {
 
-	// list of regular expressions for blocked or bad container image tags
+	// list of regular expressions specifying pattern(s) for container images that will be blocked. */
 	BlockList []string
+
+	// list of regular expressions specifying pattern(s) for container images that will be allowed.
+	AllowList []string
 }
--- a/pkg/templates/latesttag/template.go
+++ b/pkg/templates/latesttag/template.go
@@ -23,7 +23,7 @@ func init() {
 	templates.Register(check.Template{
 		HumanName:   "Latest Tag",
 		Key:         templateKey,
-		Description: "Flag applications running containers with floating container image tag, \"latest\"",
+		Description: "Flag applications running container images that do not satisfies \"allowList\" & \"blockList\" parameters criteria.",
 		SupportedObjectKinds: config.ObjectKindsDesc{
 			ObjectKinds: []string{objectkinds.DeploymentLike},
 		},
@@ -40,6 +40,20 @@ func init() {
 				blockedRegexes = append(blockedRegexes, rg)
 			}
 
+			allowedRegexes := make([]*regexp.Regexp, 0, len(p.AllowList))
+			for _, res := range p.AllowList {
+				rg, err := regexp.Compile(res)
+				if err != nil {
+					return nil, errors.Wrapf(err, "invalid regex %s", res)
+				}
+				allowedRegexes = append(allowedRegexes, rg)
+			}
+
+			if len(blockedRegexes) > 0 && len(allowedRegexes) > 0 {
+				err := fmt.Errorf("check has both \"allowList\" & \"blockList\" parameter's values set")
+				return nil, errors.Wrapf(err, "Only one of the paramater lists can be used at a time.")
+			}
+
 			return func(_ lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
 				podSpec, found := extract.PodSpec(object.K8sObject)
 				if !found {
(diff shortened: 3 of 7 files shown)


pkg/templates/latesttag/template.go
```diff
@@ -40,6 +40,20 @@ func init() {
...
+			allowedRegexes := make([]*regexp.Regexp, 0, len(p.AllowList))
+			for _, res := range p.AllowList {
+				rg, err := regexp.Compile(res)
+				if err != nil {
+					return nil, errors.Wrapf(err, "invalid regex %s", res)
+				}
+				allowedRegexes = append(allowedRegexes, rg)
+			}
+
+			if len(blockedRegexes) > 0 && len(allowedRegexes) > 0 {
```
viswajithiii, line 52: Why this constraint? We can allow both to be set, and allow `blocked` to take precedence if something is both in allowed and blocked? The following semantics seems reasonable to me:
1. If allowed is empty, then everything is allowed automatically unless it is blocked.
2. If allowed is not empty, then only things that match an allowedRegex are allowed. Even for those things, if it matches a regex in blocked, then it is not allowed.
3. If blocked is empty, then the decision is entirely determined based on the allowedRegexes.
4. If blocked is not empty, then anything that matches a blocked regex is blocked, even if it's explicitly allowed.

Priyankasaggu11929: I've improved the check to implement the above suggested semantics. Thank you for pointing!

pkg/builtinchecks/yamls/latest-tag.yaml
```diff
@@ -1,9 +1,10 @@
...
+description: "Indicates when a deployment-like object is running a container with an invalid container image"
+remediation: "Use a container image with a proper image tag satisfying the \"AllowList\" & \"BlockList\" regex patterns."
 scope:
   objectKinds:
     - DeploymentLike
 template: "latest-tag"
 params:
-  BlockList: [".*:(latest)$" ]
+  BlockList: [".*:(latest)$"]
+  AllowList: []
```
viswajithiii, line 10: We can just omit the `AllowList` if it's empty. Otherwise, it can be confusing.

pkg/templates/latesttag/template.go
```diff
@@ -49,14 +58,27 @@ func init() {
 				var results []diagnostic.Diagnostic
 
 				for _, container := range podSpec.Containers {
-					if isInList(blockedRegexes, container.Image) {
-						results = append(results, diagnostic.Diagnostic{Message: fmt.Sprintf("The container %q is using a floating image tag, %q.", container.Name, container.Image)})
-					}
+					if !(len(blockedRegexes) == 0 && len(allowedRegexes) == 0) {
```
viswajithiii, line 61: This `if` is unnecessary.

pkg/templates/latesttag/template.go
```diff
@@ -49,14 +58,27 @@ func init() {
...
+						case len(blockedRegexes) > 0 && len(allowedRegexes) == 0:
+							if isInList(blockedRegexes, container.Image) {
+								results = append(results, diagnostic.Diagnostic{Message: fmt.Sprintf("The container %q is using an invalid container image, %q. Please use images that are not blocked by the `BlockList` criteria : %q", container.Name, container.Image, blockedRegexes)})
+							}
+						case len(blockedRegexes) == 0 && len(allowedRegexes) > 0:
+							if !isInList(allowedRegexes, container.Image) {
+								results = append(results, diagnostic.Diagnostic{Message: fmt.Sprintf("The container %q is using an invalid container image, %q. Please use images that satisfies the `AllowList` criteria : %q", container.Name, container.Image, allowedRegexes)})
+							}
+						}
+					}
```
viswajithiii, line 79: I think this can be simplified to:

```go
if len(blockedRegexes) > 0 && isInList(blockedRegexes, container.Image) { // ... error}
else if len(allowedRegexes) > 0 && !isInList(allowedRegexes, container.Image) { // ... error }
```

pkg/templates/latesttag/internal/params/params.go
```diff
@@ -5,4 +5,7 @@ type Params struct {
 
 	// list of regular expressions for blocked or bad container image tags
 	BlockList []string
+
+	// list of regular expressions for allowed container image conventions
+	AllowList []string
```
viswajithiii, line 10: Please update the comments to specify what happens if allowed and blocked are both specified (ie, blocked takes precedence), and what happens if `AllowList` is empty (that it means everything is allowed)

pkg/builtinchecks/yamls/latest-tag.yaml
Priyankasaggu11929: Removed the empty `AllowList`. 

Thanks for pointing.

pkg/templates/latesttag/template.go
Priyankasaggu11929: Removed the block.

pkg/templates/latesttag/template.go
Priyankasaggu11929: Simplified the above block as suggested. Thanks for pointing.

pkg/templates/latesttag/internal/params/params.go
Priyankasaggu11929: Added the above specifications in the check description. 

Looks like [following]( https://github.com/stackrox/kube-linter/blob/c65dd74013ef482ed9ff2f8fcce4dbf926a74fae/docs/generated/templates.md#latest-tag):

Screenshot from 2021-08-12 07-20-52

pkg/templates/latesttag/template.go
rrati: In the original issue we discussed this and decided to have only 1 list to be set at at time to avoid unintentional issues with config.  There should be nothing wrong with users defining multiple checks for allow and deny vs putting in 1 check.  This alleviates any potential issues with config and allows the check to be single focused and should make it less error prone.

viswajithiii: Hmm I prefer this because it makes the configuration more readable. Often a user will have an intent like "allow only images from quay.io/my-org-name but do _not_ allow quay.io/my-org-name/image-name-used-for-dev". It would be nice to capture that in one check. This approach of having both an allowlist and denylist, and having one take precedence, is pretty common too -- for example, Kubernetes does this with add and drop capabilities.

viswajithiii (approved):

viswajithiii:
Thanks for the contribution, and sorry for the late review as I was on vacation last week!

PR #202, opened 2021-08-18
add imagePullPolicy check

glehmann:
There doesn't seem to be a consensus on what is the "good" value,
so I've only kept the check as a template.

It's meant to be used this way:

~~~yaml
customChecks:
  - name: not-always-image-pull-policy
    template: image-pull-policy
    params:
      forbiddenPolicies:
        - Always
        - Never
~~~


pkg/templates/imagepullpolicy/template.go
```diff
@@ -0,0 +1,39 @@
...
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
+				forbiddenPolicies := map[string]bool{}
+				for _, v := range p.ForbiddenPolicies {
+					forbiddenPolicies[v] = true
+				}
```
viswajithiii, line 31: This should be done inside the outer `func(p params.Params) (check.Func, error)` (ie, in line 27), so that it only has to be done once. Also, you can set the `set.StringSet` utility (from the "internal/set" folder)

pkg/templates/imagepullpolicy/template.go
```diff
@@ -0,0 +1,39 @@
...
+				if _, ok := forbiddenPolicies[string(container.ImagePullPolicy)]; ok {
+					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("container %q has a %s pull image policy", container.Name, container.ImagePullPolicy)}}
```
viswajithiii, line 33:
```suggestion
					return []diagnostic.Diagnostic{{Message: fmt.Sprintf("container %q has ImagePullPolicy set to %s", container.Name, container.ImagePullPolicy)}}
```

pkg/templates/imagepullpolicy/internal/params/params.go
```diff
@@ -0,0 +1,7 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
+	// list of bad image pull policy
```
viswajithiii, line 5: Add comment lines saying `// +noregex` and `// +notnegatable` to signal that we're not allowing regexes and negations here.

pkg/templates/imagepullpolicy/template.go
```diff
@@ -0,0 +1,39 @@
+package livenessprobe
```
viswajithiii, line 1: Could you write unit tests for this? Refer #199 for an example.

viswajithiii (commented):
Thanks for doing this!

pkg/templates/imagepullpolicy/template.go
glehmann: sure!

glehmann:
A few things I'm not sure how to do:
* how do I test that a container does *not* raise a diagnostic?
* is it possible to add some enum values for the `forbiddenPolicies` param? I tried to add `// +enum=Always` for example, but it generates some non compilable code

viswajithiii:
> A few things I'm not sure how to do:
> 
> * how do I test that a container does _not_ raise a diagnostic?
This automatically happens (the test case is supposed to be comprehensive, so if something not mentioned in the test case is raised by the check, the test will cail)
> * ~is it possible to add some enum values for the `forbiddenPolicies` param? I tried to add `// +enum=Always` for example, but it generates some non compilable code~ I've modified the validation template to support that

pkg/templates/imagepullpolicy/template.go
```diff
@@ -0,0 +1,43 @@
...
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			forbiddenPolicies := map[string]bool{}
+			for _, v := range p.ForbiddenPolicies {
+				forbiddenPolicies[v] = true
+			}
```
viswajithiii, line 34:
```suggestion
			forbiddenPolicies := set.NewStringSet(p.ForbiddenPolicies...)
```

pkg/templates/imagepullpolicy/template.go
```diff
@@ -0,0 +1,43 @@
...
+			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
+				if _, ok := forbiddenPolicies[string(container.ImagePullPolicy)]; ok {
```
viswajithiii, line 36:
```suggestion
				if forbiddenPolicies.Contains(string(container.ImagePullPolicy)) {
```

pkg/templates/imagepullpolicy/template.go
```diff
@@ -0,0 +1,43 @@
+package imagepullpolicy
+
+import (
+	"fmt"
+
+	"golang.stackrox.io/kube-linter/pkg/check"
```
viswajithiii, line 6:
```suggestion
	"golang.stackrox.io/kube-linter/internal/set"
	"golang.stackrox.io/kube-linter/pkg/check"
```

viswajithiii (approved):

viswajithiii:
Thanks @glehmann!

PR #203, opened 2021-08-19
WIP: add --values flag to pass extra values files to helm

### LINKED ISSUE #197: [FEATURE_REQUEST] support dynamic helm values
I have a helm chart that I'd like to lint with your magical tool. The only problem is that I haven't found a way to dynamically pass in values with the kubelinter cli.

For example, this is how I generally invoke the helm linter: 
`helm lint myhelm -f my-dynamic-values.yaml --strict`   -- The flag  `-f my-dynamic-values.yaml` is a short yaml file of values that must be set in order for the chart to succeed in building. If those values are not properly set, my helm chart is designed to fail. This is a somewhat common helm pattern/design.

If applicable, please paste in the existing KubeLinter output along with the input used, and point out which part should be modified (expected output).
n/a

**Additional context**
I chatted with Viswajith Venugopal on slack about this feature.

glehmann:
I'm quite sure you won't like the mix of standard and variadic
arguments at various positions (I wouldn't). This is a first try, to be able to
discuss how to implement that feature.

should close #197

### DIFF
--- a/pkg/command/lint/command.go
+++ b/pkg/command/lint/command.go
@@ -76,7 +76,7 @@ func Command() *cobra.Command {
 				fmt.Fprintln(os.Stderr, "Warning: no checks enabled.")
 				return nil
 			}
-			lintCtxs, err := lintcontext.CreateContexts(args...)
+			lintCtxs, err := lintcontext.CreateContexts(cfg.Checks.Values, args...)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


viswajithiii:
Thanks @glehmann. Give me a day or two to think about this. I feel like this should be implemented slightly differently, and I will need to try and understand what make sense.

cpvandehey:
Any update here?

glehmann:
Just closing my outdated pull requests.

PR #204, opened 2021-08-19
add match to envvar

siperman:


viswajithiii (commented):
I'm a bit confused. Can you explain the use-case for this `Match` parameter?

siperman:
use-case:
Different teams need to check whether the format of environment variables is uniform.
It is also necessary to check that the environment variable conforms to a regular expression.

eg:
AppDynamics app name:  ^WEALTH\-(\w{2})\-(\w{4})\-[\w_]+\-(DEV|UAT|PROD)$

viswajithiii:
I understand that. But can't this be achieved under the status quo? How does this new parameter help?

siperman:
Because my test reverses the entire regular, it does not take effect. Remember that the reason for the search is that the golang regular library does not support inversion.

viswajithiii:
> Because my test reverses the entire regular, it does not take effect. Remember that the reason for the search is that the golang regular library does not support inversion.

Ohh I see. Yeah, we currently support this by allowing people to add a `!` at the beginning, to signify negation. This is probably not as well-documented as it should be.

viswajithiii:
@siperman You should be able to set the value in the check to `!^WEALTH-(\w{2})-(\w{4})-[\w_]+-(DEV|UAT|PROD)$`, and then I think it will do what you want?

janisz:
@siperman What's the status of this?

janisz:
Closing due to no activity on this.
If you like to work on this please reopen or create a new PR.

PR #209, opened 2021-09-13
Return non-zero exit status on no valid objects

ytsarev:
Instead of printing the warning and return 0 exit status
return Error and exit with 1

Reasoning: our kube-linter pipelines where silently failing
for a while without any notice because of 0 exit status

See https://github.com/k8gb-io/k8gb/pull/615#issuecomment-918000074

Signed-off-by: Yury Tsarev <yury.tsarev@absa.africa>

### DIFF
--- a/pkg/command/lint/command.go
+++ b/pkg/command/lint/command.go
@@ -95,8 +95,7 @@ func Command() *cobra.Command {
 				}
 			}
 			if !atLeastOneObjectFound {
-				fmt.Fprintln(os.Stderr, "Warning: no valid objects found.")
-				return nil
+				return errors.New("no valid objects found")
 			}
 			result, err := run.Run(lintCtxs, checkRegistry, enabledChecks)
 			if err != nil {
--- a/pkg/command/lint/command.go
+++ b/pkg/command/lint/command.go
@@ -44,6 +44,7 @@ var (
 // Command is the command for the lint command.
 func Command() *cobra.Command {
 	var configPath string
+	var failIfNoObjects bool
 	var verbose bool
 	format := flagutil.NewEnumFlag("Output format", formatters.GetEnabledFormatters(), common.PlainFormat)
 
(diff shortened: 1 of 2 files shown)


pkg/command/lint/command.go
```diff
@@ -95,8 +95,7 @@ func Command() *cobra.Command {
 				}
 			}
 			if !atLeastOneObjectFound {
-				fmt.Fprintln(os.Stderr, "Warning: no valid objects found.")
-				return nil
```
viswajithiii, line 99: This breaks backward compatibility, and is more in line with how I'd expect a linter to work -- just emit a warning if there are no valid objects, but don't fail altogether. I would be open to adding a flag, though, that controls this, so that it someone passes it in (say, "--fail-if-no-objects-found"), _then_ we return an error. What do you think?

viswajithiii (commented):
Thanks for the PR!

jkremser:
@viswajithiii I've added the `--fail-if-no-objects-found` flag and some testing, could you please take a look? The CI is failing on something unrelated.

I've run the new e2e tests locally by

```bash
KUBE_LINTER_BIN=~/workspace/kube-linter/bin/darwin/kube-linter go test e2etests/sanity_test.go
```
..and they passed.

We use the `kube-linter` via gh action and It would be nice to have this in upstream, otherwise we have to do some ugly workarounds like [this](https://github.com/jkremser/k8gb/blob/d1043c84f8e90b6e4681986313cf87f2d6c30e13/.github/workflows/kube-linter.yaml#L22:L24).

jkremser:
rebased on `main` with the hope that the CI will pass

ytsarev:
@viswajithiii hey, how does it look to you now? I think @jkremser addressed the issue the best possible way

viswajithiii (approved):
Thank you! Sorry for the late review, I missed the previous notification.

jkremser:
@viswajithiii thanks for merging :+1:

 may I ask if there will be a release with the change? Or if there is a way to specify the version of the `kube-linter` tool from `kube-linter-action` to something like gitsha or branch/tag.  It looks like the version can be set to `latest`, but this [means](https://github.com/stackrox/kube-linter-action/blob/main/action.yml#L37:L40) the "latest released", right?

janisz:
Sure I can handle this after the weekend

Issue #212, opened 2021-09-19
[BUG] Wrong artifacts locations in SARIF reports for Helm if Chart and containing directory have different names


rm3l:
**System info:**
 ```
 ❯ kube-linter version
development
```
 - Ubuntu Linux 21.04

**Describe the bug**
Hi there!
I was trying to run KubeLinter against a set of Helm Charts and upload the generated SARIF reports to GitHub Code Scanning.
I noticed the artifacts locations in the generated SARIF reports use the Helm Chart names in the relative paths, which obviously could not be resolved by GitHub Code Scanning because the Charts live in directories named differently.
Maybe I missed something, but as mentioned in https://github.com/helm/helm/issues/1979 and [helm/helm-www#113](https://github.com/helm/helm-www/pull/1113/files#diff-c31785a9696cc3162b9e6837ee45b28b5eaa978f62970adecd6c2e294d5b7ea8), this constraint on the directory name has been removed in Helm and is no longer a requirement for Helm Charts.

**To Reproduce**
Run the following commands:

```sh
# Create a sample Helm Chart
helm create my-helm-chart

# Rename the containing directory
mv my-helm-chart my-helm-chart-directory

# kube-linter
kube-linter lint my-helm-chart-directory --format sarif \
  | jq -r '.runs[].results[].locations[].physicalLocation.artifactLocation.uri' \
  | uniq
```

The `kube-linter` command above returns the following relative paths, which cannot be resolved by tools like GitHub Code Scanning.
```
my-helm-chart/templates/deployment.yaml
my-helm-chart/templates/tests/test-connection.yaml
```

**Sample YAML input**
N/A

**Expected behavior**
I would expect the artifact locations paths to use the containing directory. In the sample commands above, I would expect the following paths instead:
```
my-helm-chart-directory/templates/deployment.yaml
my-helm-chart-directory/templates/tests/test-connection.yaml
```

**Additional context**
Below is a screen capture illustrating how GitHub Code Scanning is unable to display a preview of the YAML resources, because it cannot resolve the paths reported in the SARIF reports:
![Screenshot from 2021-09-20 00-07-11](https://user-images.githubusercontent.com/593208/133944574-3d08f808-3c3d-4e15-9563-534d1f91e883.png)

msugakov:
Thanks again for the detailed description and steps to reproduce, Armel!
I published today a PR to address this issue https://github.com/stackrox/kube-linter/pull/215
Will let you know once it's released.

[...]

PR #215, opened 2021-09-21
Fix helm paths, see issue #212

msugakov:
See https://github.com/stackrox/kube-linter/issues/212 motivating this change.

Note that here we have a bit of modifications not related to each other. I tried to make all commits self-described.
Therefore, it is **best to review** this change [**by individual commits**](https://github.com/stackrox/kube-linter/pull/215/commits) while looking at commit messages explaining why or how for each commit.

Testing:
* Relying on CI.
* Verified on the example from the issue #212 
```bash
$ make build
$ .gobin/kube-linter lint my-helm-chart-directory --format sarif \
  | jq -r '.runs[].results[].locations[].physicalLocation.artifactLocation.uri'
Error: found 12 lint errors
my-helm-chart-directory/templates/tests/test-connection.yaml
my-helm-chart-directory/templates/tests/test-connection.yaml
my-helm-chart-directory/templates/tests/test-connection.yaml
my-helm-chart-directory/templates/tests/test-connection.yaml
my-helm-chart-directory/templates/tests/test-connection.yaml
my-helm-chart-directory/templates/tests/test-connection.yaml
my-helm-chart-directory/templates/deployment.yaml
my-helm-chart-directory/templates/deployment.yaml
my-helm-chart-directory/templates/deployment.yaml
my-helm-chart-directory/templates/deployment.yaml
my-helm-chart-directory/templates/deployment.yaml
my-helm-chart-directory/templates/deployment.yaml
```


pkg/templates/updateconfig/template_test.go
```diff
@@ -499,8 +508,14 @@ func (s *UpgradeConfigTestSuite) TestMinSurgeInteger() {
 				},
 				deploymentWithValidRollingUpdate:       {},
 				deploymentConfigWithValidRollingUpdate: {},
-				daemonSetWithStrategy:                  {},
-				replicationController:                  {},
+				daemonSetInvalidStrategy: {
+					{Message: "object has a max surge of <nil> but at least 2 is required"},
```
msugakov, line 512: @rrati Here's the case we discussed in CoreOS slack. Formerly the test was called `daemonSetWithStrategy`. I renamed it to `daemonSetInvalidStrategy`.
It used to expect no message `{}`, which, as we discussed, is wrong. Now it expects the message and the test succeeds after dependency updates from the previous commit [`7148bed` (#215)](https://github.com/stackrox/kube-linter/pull/215/commits/7148bed15124a12bdc75522b0258e66b571a5571). (I am not curious enough to trace down which update fixed DaemonSet behavior.)

pkg/lintcontext/create_contexts.go
```diff
@@ -113,13 +112,3 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 	}
 	return contexts, nil
 }
-
-// CreateContextsFromHelmArchive creates a context from TGZ reader of Helm Chart.
-func CreateContextsFromHelmArchive(fileName string, tgzReader io.Reader) ([]LintContext, error) {
```
viswajithiii, line 118: We can't remove this function, it was added for library use and removing this will break code that uses it. See https://github.com/stackrox/kube-linter/pull/173

viswajithiii (approved):
This looks great, Misha, thanks for doing it!

Issue #217, opened 2021-09-22
[FEATURE_REQUEST] 'deprecated-service-account-field' should pass if both 'serviceAccount' and 'serviceAccountName' are set


bkez322:
**Description of the problem/feature request**
[The current implementation](https://github.com/stackrox/kube-linter/blob/c12af3736628d22ea6b84a1876ac4ca01174c57a/pkg/templates/deprecatedserviceaccount/template.go#L32-L35) of the 'deprecated-service-account-field' check is to fail the check if a pod-spec has a non-empty `DeprecatedServiceAccount`(yaml: `serviceAccount`) property.

However, I propose it should be considered a valid use-case for a deployment resource YAML to have both  `serviceAccount` and `serviceAccountName` set (with the same value) for backward compatibility purposes. 

I propose that kube-linter should pass the ‘deprecated-service-account-field’ check if either one of the following conditions are true:
* `serviceAccount` not set at all (this is the only valid condition currently)
* `serviceAccount` and `serviceAccountName` are both set, and have matching values

**Existing Behavior**
```yml
# nginx-deployment.yml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
      serviceAccount: my-service-account
      serviceAccountName: my-service-account
```
`kube-linter lint nginx-deployment.yml` will fail with the following error:
```
nginx-deployment.yml: (object: default/nginx-deployment apps/v1, Kind=Deployment) serviceAccount is specified (my-service-account), but this field is deprecated; use serviceAccountName instead (check: deprecated-service-account-field, remediation: Use the serviceAccountName field instead.)
```

**Expected Behavior**
The YAML above should not result in a failed ‘deprecated-service-account-field’ check.

**Additional Context**
It may be worth noting that kubernetes [will automatically add](https://github.com/kubernetes/kubernetes/blob/5b489e2846a7fb959252dc5a04fe21ec844e9fad/pkg/apis/core/v1/conversion.go#L297) `serviceAccount` to a post-apply in-cluster deployment resource YAML if `serviceAccountName` is present, to promote backward compatibility.

viswajithiii:
[...]

PR #218, opened 2021-09-22
'deprecated-service-account-field' should pass if both 'serviceAccount' and 'serviceAccountName' are set

bkez322:
Resolves: #217 

Updates the 'deprecated-service-account-field' check to pass if either of the two conditions are met:
* serviceAccount not set at all (this is the only valid condition prior to this PR)
* serviceAccount and serviceAccountName are both set, and have matching values (new condition)

Output messages are updated as appropriate.


viswajithiii (approved):
LGTM, thanks!

PR #219, opened 2021-09-23
Correctly detect the replica count and selector for DeploymentConfig objects

### LINKED ISSUE #216: [BUG] Replicas doesnt work for DeploymentConfigs
**System info:**
 - OS: MaxOS

**Describe the bug**
If you switch the replicas tests for deploymentconfigs, you get nothing back.
- https://github.com/stackrox/kube-linter/blob/main/pkg/templates/replicas/template.go#L33

It's because of a type mismatch:
- https://github.com/stackrox/kube-linter/blob/c12af3736628d22ea6b84a1876ac4ca01174c57a/pkg/extract/pod_spec.go#L87

**To Reproduce**
Apply attached patch, run: `replicas` tests only

**Expected behavior**
Works for a DeploymentConfig

garethahealy:
Fixes #216 and #220

### DIFF
--- a/pkg/extract/pod_spec.go
+++ b/pkg/extract/pod_spec.go
@@ -84,6 +84,9 @@ func Replicas(obj k8sutil.Object) (int32, bool) {
 	if !replicas.IsValid() {
 		return 0, false
 	}
+
+	//TODO: below is needed for DeploymentConfigs
+	//numReplicas, ok := replicas.Interface().(int32)
 	numReplicas, ok := replicas.Interface().(*int32)
 	if ok {
 		if numReplicas != nil {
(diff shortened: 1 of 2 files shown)


pkg/extract/pod_spec.go
```diff
@@ -84,6 +84,9 @@ func Replicas(obj k8sutil.Object) (int32, bool) {
 	if !replicas.IsValid() {
 		return 0, false
 	}
+
+	//TODO: below is needed for DeploymentConfigs
+	//numReplicas, ok := replicas.Interface().(int32)
```
garethahealy, line 89: work required to get the tests to pass

viswajithiii (approved):

PR #230, opened 2021-09-30
Ports check: Assume protocol is TCP if not specified

### LINKED ISSUE #221: [BUG] Default pod ports protocol
**System info:**
 - OS: MaxOS

**Describe the bug**
The `port.Protocol` doesn't default to anything currently:
- https://github.com/stackrox/kube-linter/blob/main/pkg/templates/ports/template.go#L37

which means if you run the `ssh-port` check but the dev hasn't included the protocol stanza, this check doesn't fail

**To Reproduce**
kube-linter lint --include ssh-port --do-not-auto-add-defaults {yaml-provided.yml}

**Sample YAML input**
```
apiVersion: apps.openshift.io/v1
kind: DeploymentConfig
metadata:
  name: app2
spec:
  replicas: 3
  template:
    spec:
      containers:
      - name: app
        ports:
        - containerPort: 22
```

**Expected behavior**
As TCP is the default option, should it default in the template if its nil
- https://kubernetes.io/docs/concepts/services-networking/service/#protocol-support

viswajithiii:
Fixes #221

### DIFF
--- a/pkg/templates/ports/template.go
+++ b/pkg/templates/ports/template.go
@@ -34,9 +34,15 @@ func init() {
 			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
 				var results []diagnostic.Diagnostic
 				for _, port := range container.Ports {
-					if int(port.ContainerPort) == p.Port && protocolMatcher(string(port.Protocol)) {
+					// The k8s protocol defaults to TCP even if not set in the YAML.
+					protocol := string(port.Protocol)
+					if protocol == "" {
+						protocol = "TCP"
+					}
+					if int(port.ContainerPort) == p.Port && protocolMatcher(protocol) {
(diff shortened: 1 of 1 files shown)


md2119 (approved):

pkg/templates/ports/template.go
```diff
@@ -34,9 +34,15 @@ func init() {
 			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
 				var results []diagnostic.Diagnostic
 				for _, port := range container.Ports {
-					if int(port.ContainerPort) == p.Port && protocolMatcher(string(port.Protocol)) {
+					// The k8s protocol defaults to TCP even if not set in the YAML.
```
janisz, line 37: It will be great to add a test for this change

Issue #232, opened 2021-10-13
latest-tag rule does not catch no-tag


hassenius:
The builtin [latest-tag rule](https://github.com/stackrox/kube-linter/blob/main/pkg/builtinchecks/yamls/latest-tag.yaml) does not catch images with no tag (which will default to :latest by the pulling client), hence get past the validation.

PR #233, opened 2021-10-13
WIP: add check for no tag as latest

hassenius:
As discussed in #232 , no tag is interpreted as `latest` when pulling, so should also fail linter.

These regexes have purely been worked out over at https://regex101.com

Could not see a simple way to cater for both the cases where you have

```
image: nginx
```
(matched by `^((?!:).)*$`)
and 

```
image: myownhost:5000/myrepo/nginx
```
(matched by `^(.*\/[^:]+)$`)

### DIFF
--- a/.github/workflows/test-sarif.yaml
+++ b/.github/workflows/test-sarif.yaml
@@ -65,7 +65,7 @@ jobs:
           GO111MODULE=on go get github.com/neilpa/yajsv@v1.4.0
           schema=$(jq -r '.["$schema"]' results.sarif)
           [ "$schema" = https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json ]
-          yajsv -s ./scripts/sarif-schema-2.1.0.json results.sarif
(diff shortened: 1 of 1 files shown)


viswajithiii:
Thanks for sending this over @hassenius -- it looks like we're going to have to change the regex slightly to match Golang format (Perl-style doesn't work)

.github/workflows/test-sarif.yaml
```diff
@@ -64,8 +64,8 @@ jobs:
           set -ex
           GO111MODULE=on go get github.com/neilpa/yajsv@v1.4.0
           schema=$(jq -r '.["$schema"]' results.sarif)
-          [ -n "$schema" ] && wget -nv $schema
-          yajsv -s $(basename $schema) results.sarif
+          [ "$schema" = https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json ]
```
janisz, line 67: I think we should not use `master` here for now this link returns 404
We should always use links with commit hash.

PR #236, opened 2021-10-25
WIP: Add check for too many of a kind

npecka:
Adding check for DVO-9. Still WIP. Need to put in method for check to keep track of kinds and only alert when it sees a duplicate of known kinds within cluster


janisz:
@npecka What's the status of this?

janisz:
Stale, closing

PR #237, opened 2021-10-26
WIP: Don't report metric if replica count is 0

npecka:
DVO-7


viswajithiii:
Hi @npecka, can you explain the reasoning behind this?

PR #238, opened 2021-10-28
added integration tests via bats-core

garethahealy:
One thing which I found when playing with kube-linter was I couldn't find many tests which showed the resources which a check would validate against. this PR, adds those examples as tests via `bats`

The basic idea is: for every built-in check, there is a bats test that runs the kube-linter process with only that check enabled. the YAML input has 2 examples (_in most cases_), one that shouldn't trigger the check and one that should.

to run the bats tests, you just do `make bats`. if `diff` is on the machine, it'll also compare the tests to what the built-in checks list outputs to cover off anything missed/removed/new etc. and finally, that is run via a github action.

### DIFF
--- a/e2etests/bats-tests.sh
+++ b/e2etests/bats-tests.sh
@@ -1,0 +1,726 @@
+#!/usr/bin/env bats
+
+load bats-support-clone
+load test_helper/bats-support/load
+load test_helper/redhatcop-bats-library/src/error-handling.bash
+
+# NOTE: Each test matches to a built-in check outputted via 'kube-linter check'
+
+get_value_from() {
+  value=$(echo "${1}" | jq -r "${2}")
+
+  if [[ -z "${value}" ]] ; then
+    fail "# FATAL-ERROR: get_value_from: value is empty or invalid" || return $?
+  fi
+
+  echo "${value}"
+}
+
+@test "access-to-create-pods" {
+  tmp="tests/checks/access-to-create-pods.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include access-to-create-pods --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+
+  message1=$(get_value_from "${lines[0]}" '.Reports[0].Object.K8sObject.GroupVersionKind.Kind + ": " + .Reports[0].Diagnostic.Message')
+  count=$(get_value_from "${lines[0]}" '.Reports | length')
+
+  [[ "${message1}" == "RoleBinding: binding to \"role1\" role that has [create] access to [pods]" ]]
+  [[ "${count}" == "1" ]]
+}
+
+@test "access-to-secrets" {
+  tmp="tests/checks/access-to-secrets.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include access-to-secrets --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+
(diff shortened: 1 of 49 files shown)


e2etests/bats-tests.sh
```diff
@@ -0,0 +1,527 @@
...
+  tmp="tests/checks/access-to-create-pods.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include access-to-create-pods --do-not-auto-add-defaults ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+  [[ "${lines[1]}" == "tests/checks/access-to-create-pods.yml: (object: namespace-dev/role-binding1 rbac.authorization.k8s.io/v1, Kind=RoleBinding) binding to \"role1\" role that has [create] access to [pods] (check: access-to-create-pods, remediation: Where possible, remove create access to pod objects in the cluster.)" ]]
+  [[ "${lines[2]}" == "Error: found 1 lint errors" ]]
+  [[ "${#lines[@]}" -eq 3 ]]
+}
```
viswajithiii, line 17: There is a lot of boilerplate here. I wonder if it would be possible to specify this more declaratively somehow, versus requiring all this to be copy pasted?

e2etests/bats-tests.sh
```diff
@@ -0,0 +1,527 @@
...
+load test_helper/redhatcop-bats-library/src/error-handling.bash
+
+@test "access-to-create-pods" {
+  tmp="tests/checks/access-to-create-pods.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include access-to-create-pods --do-not-auto-add-defaults ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+  [[ "${lines[1]}" == "tests/checks/access-to-create-pods.yml: (object: namespace-dev/role-binding1 rbac.authorization.k8s.io/v1, Kind=RoleBinding) binding to \"role1\" role that has [create] access to [pods] (check: access-to-create-pods, remediation: Where possible, remove create access to pod objects in the cluster.)" ]]
```
viswajithiii, line 14: I think we should use the JSON format, and use `jq` to do the assertions. Otherwise, if we change the output format, updating the tests will be a pain. The tests would also be more readable that way.

viswajithiii (commented):
Hi @garethahealy, this looks amazing, thanks so much for the effort you've put in to do this.

Some instructions on how to run this test locally would be helpful. We should add those instructions to a README folder in the repo. We should also add instructions on what someone needs to do to add a new test case. I have actually never used `bats` before, and if I naively run `make bats`, this is what I see:

```
)$ make bats
KUBE_LINTER_BIN="/Users/viswa/development/src/github.com/stackrox/kube-linter/.gobin/kube-linter" e2etests/bats-tests.sh
env: bats: No such file or directory
make: *** [bats] Error 127
```

e2etests/bats-tests.sh
garethahealy: pushed a commit (https://github.com/stackrox/kube-linter/pull/238/commits/560d5615c80ba0c651405a01bff485be66adfe3c) - do you mean something like this?

e2etests/bats-tests.sh
garethahealy: I can't think of a less copy+paste way than what I've currently done, but open to suggestions/ideas

e2etests/bats-tests.sh
viswajithiii: Yes, exactly.

e2etests/bats-tests.sh
viswajithiii: Fair enough. Looking at it again, boilerplate is quite minimal.

garethahealy:
@viswajithiii ; pushed requested changes

viswajithiii (approved):
Thanks, this is really helpful!

PR #246, opened 2021-12-16
control registering open-shift kinds with a flag

deepak1725:
While Running `kube-linter` in an Operator for Dynamic Analyses in a Kubernetes server, Openshift kinds were Not recognised.

Idea is to control registering OpenShift kinds with a flag.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -31,8 +31,9 @@ type ChecksConfig struct {
 // Config represents the config file format.
 type Config struct {
 	// +flagName=-
-	CustomChecks []Check      `json:"customChecks,omitempty"`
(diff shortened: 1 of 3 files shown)


viswajithiii (commented):
Hey @deepak1725, thanks for filing this PR. I'm not sure I follow though. Why would you not want to allow openshift kinds?

deepak1725:
We are extending kube-linter for Dynamic runtime analyses. 
So, in native Kube envs, types like `DeploymentConfig` creating a CrashLoop Backoff. It is possible if we can control this via flag? Any default values(true/false) would be okay. (We would override it, by detecting user env.)

PR #247, opened 2021-12-23
docs: Use YAML for parameters in documentation

janisz:
Currently we display checks and template params as JSON without highlighting (see #192). What's more for checks we use plain JSON and for templates pretty printed. This is weird as in our own codebase and examples we use YAML. 

This PR:
- changes JSON to YAML in generated documentation 
- changes JSON to YAML in `[checks|templates] list --format markdown` output
- remove `Parameters` section when there is no parameters  

Closes: #192 as we don't have JSON in our docs anymore

### DIFF
--- a/docs/generated/checks.md
+++ b/docs/generated/checks.md
@@ -14,10 +14,18 @@ KubeLinter includes the following built-in checks:
 
 **Parameters**:
 
-```json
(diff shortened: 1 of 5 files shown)


viswajithiii (approved):
Good idea. The JSON is from back when I tried to render all of these in cells within a table, but now that they are free form this makes a lot more sense

Issue #244, opened 2021-12-14
[FEATURE_REQUEST] add support to check certain annotations are not used


sagar-shah-uipath:
**Description of the problem/feature request**
We wanna make sure that some developers don't use certain annotations in code base because that can cause deployments to restart.
E.g - Stakater provides `reloader.stakater.com/auto`, whenever any deployment/job has that annotation attached to it, they get restarted whenever any config or secret gets changed in namespace. In our use case, we wanna make sure nobody uses `reloader.stakater.com/auto` annotation bcz otherwise it will keep on restarting our deployments if config gets changed frequently

Request is similar to required-annotations, but what we need here different is this annotation should not be present anywhere in any resources.

viswajithiii:
Thanks @sagar-shah-uipath for filing this issue! I've added it to our roadmap.

PR #248, opened 2021-12-23
template: add forbidden-annotation

janisz:
Fixes: #244


pkg/templates/forbiddenannotation/internal/params/params.go
```diff
@@ -0,0 +1,12 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
+
+	// Key of the forbidden label.
```
viswajithiii, line 6:
```suggestion
	// Key of the forbidden annotation.
```

pkg/templates/forbiddenannotation/internal/params/params.go
```diff
@@ -0,0 +1,12 @@
...
+	// +required
+	Key string
+
+	// Value of the forbidden label.
```
viswajithiii, line 10:
```suggestion
	// Value of the forbidden annotation.
```

viswajithiii (approved):
This looks great, and thanks for the thorough testing!

PR #258, opened 2022-01-24
update go-sarif to v2

owenrumney:
Relatively trivial PR to update Sarif to v2. On the face of it, it doesn't really change your usage or generation of Sarif in its current form; however, v2 no completely supports the current SARIF spec.

As its a breaking change, I'm going round the repos that use it offering to fix the break with a PR - I appreciate you may chose to stay with v1 which is cool!

### DIFF
--- a/.github/workflows/test-sarif.yaml
+++ b/.github/workflows/test-sarif.yaml
@@ -67,7 +67,7 @@ jobs:
           set -ex
           GO111MODULE=on go get github.com/neilpa/yajsv@v1.4.0
           schema=$(jq -r '.["$schema"]' results.sarif)
-          [ "$schema" = https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json ]
+          [ "$schema" = https://json.schemastore.org/sarif-2.1.0-rtm.5.json ]
           yajsv -s ./scripts/sarif/sarif-schema-2.1.0.json results.sarif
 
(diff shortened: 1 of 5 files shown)


janisz:
@owenrumney Thanks for contribution!

I take a look at our SARIF test and it looks like we have hardcoded URL to schema and that leads to failures 

https://github.com/stackrox/kube-linter/blob/40da4f40cb329edb8f4452567f187e9d752cb1ce/.github/workflows/test-sarif.yaml#L67

I think we don't need to check generated URL but only confront output with schema.

I run test manually and it's passing with schema we have have in [`/scripts/sarif/sarif-schema-2.1.0.json`](https://github.com/stackrox/kube-linter/blob/499d60f66df8be0e597eba77d64fbd2c64cca6ad/scripts/sarif/sarif-schema-2.1.0.json)

```
$ yajsv -s ./scripts/sarif/sarif-schema-2.1.0.json results.sarif
results.sarif: pass
```

owenrumney:
Microsoft alerted me to the change in the schema in https://github.com/owenrumney/go-sarif/issues/26 

I've updated the action and the schema in scripts to be the correct one and validated with the steps below

1. make build

2. bin/linux/kube-linter lint test/testdata/splunk.yaml > result.sarif

3. yajsv -s ./scripts/sarif/sarif-schema-2.1.0.json result.sarif
result.sarif: pass

janisz (approved):

janisz:
@owenrumney Thanks!

PR #279, opened 2022-03-25
Feat/fail invalid yaml

### LINKED ISSUE #272: [FEATURE_REQUEST] Fail linting if files cannot be decoded

Kube-linter will silently ignore files that it cannot decode. This makes sense for ignoring files that aren't kubernetes resources but in the case that they are, kube-linter should fail linting. 

Where I notice this the most is invalid [quantities](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/
) for resources (cpu, memory, storage, etc). I want to fail our CI built so our GitOps engine does not sync out invalid manfiests. 

When kube-linter tries to decode a kubernetes resource that is invalid, it silently ignores the resource. I suggest that we add a config option that allows these errors to fail the linting. 

**Additional context**
kubeval is also struggling with similar issues
1. https://github.com/instrumenta/kubeval/issues/185
2. https://github.com/instrumenta/kubeval/issues/195
3. https://github.com/instrumenta/kubeval/issues/208

**Example**
```
cat <<  EOF > pod-test.yaml 
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: null
  labels:
    run: priv-pod
  name: priv-pod
  namespace: default
spec:
  containers:
  - image: busybox
    name: invalid
    command:
      - "sleep"
    args:
      - "infinity"
    resources:
      limits:
        cpu: 25m
        memory: 1GB
      requests:
        cpu: 25m
        memory: 1GB
  dnsPolicy: ClusterFirst
  restartPolicy: Always
status: {}
EOF
```
Lint the file.
```
>  kube-linter lint pod-test.yaml
Warning: no valid objects found.

> kube-linter lint pod-test.yaml -v                                                          
Warning: failed to load object from pod-test.yaml: failed to decode: v1.Pod.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Resources: v1.ResourceRequirements.Requests: Limits: unmarshalerDecoder: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$', error found in #10 byte of ...|ory":"1GB"},"request|..., bigger context ...|,"resources":{"limits":{"cpu":"25m","memory":"1GB"},"requests":{"cpu":"25m","memory":"1GB"}}}],"dnsP|...
Warning: no valid objects found.
```

rumstead:
closes #272


pkg/command/lint/command.go
```diff
@@ -81,13 +82,23 @@ func Command() *cobra.Command {
...
+				}
 				for _, lintCtx := range lintCtxs {
 					for _, invalidObj := range lintCtx.InvalidObjects() {
-						fmt.Fprintf(os.Stderr, "Warning: failed to load object from %s: %v\n", invalidObj.Metadata.FilePath, invalidObj.LoadErr)
+						hasInvalidObjects = true
+						_, _ = fmt.Fprintf(os.Stderr, "%s: failed to load object from %s: %v\n", prefix, invalidObj.Metadata.FilePath, invalidObj.LoadErr)
 					}
 				}
+				if hasInvalidObjects && errorOnInvalidResource {
+					return fmt.Errorf("unable to load all objects")
```
janisz, line 98: I'm not sure if we should short circuit here. How about you have 10 files but only one is invalid. Instead of getting lint errors for all of them you'll get only information about invalid one. I think better if we add this results to other linters so we can have them in json or sarif format as well helping automation tools to mark them as problematic. 
What do you think about this? On the other hand we can start small (merge this PR) and then iterate on what we have.

rumstead: I was thinking about this a well. These errors also aren't "pretty printed". I am not sure how to pass the results down. I would need to either append them to the `result` array at end. I also would need to fill in the `Result` object. 

Happy to do it but it was a larger change. Any impl suggestions?

rumstead: @janisz, don't mean to be pushy but any thoughts here? I can definitely "hack" it into a result struct but want to make sure we agree before I do it.

janisz: I'm sorry I was on holiday with limited access to internet. I'll try to find a way how to simplify the implementation

pkg/command/lint/command_test.go
```diff
@@ -0,0 +1,45 @@
...
+func TestCommand_InvalidResources(t *testing.T) {
+	testDataPath := getTestDataDir()
+	tests := []struct {
+		name    string
+		cmd     *cobra.Command
+		failure bool
+		output  string
+	}{
+		{name: "InvalidPodResource", cmd: createLintCommand(testDataPath+"invalid-pod-resources.yaml", "--fail-on-invalid-resource"), failure: true},
+		{name: "InvalidPVCResource", cmd: createLintCommand(testDataPath+"invalid-pvc-resources.yaml", "--fail-on-invalid-resource"), failure: true},
```
janisz, line 23: How about adding missing file and not a yaml test cases?

janisz: Maybe a valid one too. We can use `pod.yaml` from homebrew https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/kube-linter.rb#L38-L58

pkg/command/lint/command_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+func createLintCommand(args ...string) *cobra.Command {
+	c := Command()
+	c.SetArgs(args)
+	return c
+}
+
+func getTestDataDir() string {
```
janisz, line 41: do we need that? When simple `testdata/` will not work?

janisz (commented):
LGTM, thanks for adding the test!
I think we may need some more test cases and I'm not sure about need for get test data dir function.

pkg/command/lint/command_test.go
rumstead: yup not required when they are in the same folder

rumstead:
Should be G2G lmk if there is anything else

rumstead:
@viswajithiii able to review this?

janisz (approved):

janisz:
@rumstead Awesome, thanks for this contribution. What's more I think this flag should be by default enabled but this is something we can handle for 1.0.0 release. It's worth to mention it in docs.

PR #296, opened 2022-04-21
Use single workflow to build and test

janisz:
Now we build kube-linter 3 times which is way too much. This PR changes workflow to build kube-linter just once and let other jobs depend on that built.


.github/workflows/build.yaml
```diff
@@ -68,6 +74,49 @@ jobs:
       - name: Run bats tests
         run: make e2e-bats
```
janisz, line 75: Bats are running here

.github/workflows/build.yaml
```diff
@@ -68,6 +68,54 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Download executable
+        uses: actions/download-artifact@v2
+        with:
+          name: bin
+      
+      - name: Set permissions to file
```
dhaus67, line 82: nit: Maybe we can merge both together into a single step, i.e. `Make kube-linter executable and verify version`?

dhaus67 (approved):

PR #302, opened 2022-04-25
Do not alert on missing readiness / liveness probe for init containers.

### LINKED ISSUE #261: [BUG] kube-linter reports violations for no-liveness-probe in initContainer
**System info:**
 - All OS

**Describe the bug**
- We came across a bug with kube-linter where kube-linter reports violations for no-liveness-probe . Ideally it should not show violations for initContainers since k8s does not support adding liveness probe for initContainers but in our case, it is reporting violations for initContainers as well

**To Reproduce**
- Run kube-linter for any deployment that has initContainer where liveness probe is not configured

**Sample YAML input**
If applicable, sample YAML input which reproduces the issue.

**Expected behavior**
- It should not show violation for initContainer
https://kube-linter.slack.com/archives/C01ASRW6Y8L/p1643282582226739

dhaus67:
Create new util function `PerNonInitContainer`, which will skip init containers from being taken into account for readiness / liveness probe checks, as [init containers cannot specify them according to the pod spec.](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#PodSpec)

Closes #261

### DIFF
--- a/pkg/templates/livenessprobe/template.go
+++ b/pkg/templates/livenessprobe/template.go
@@ -24,7 +24,7 @@ func init() {
 		Parameters:             params.ParamDescs,
 		ParseAndValidateParams: params.ParseAndValidate,
 		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
-			return util.PerContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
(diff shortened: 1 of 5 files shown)


viswajithiii (approved):
Makes sense

janisz (approved):

PR #306, opened 2022-04-27
Ignore dependabot branches.

dhaus67:
Currently dependabot branches CI is failing. We execute `test sarif` on push events which requires write access, dependabot only has read access on push events.

Skipping dependabot branches' push event to avoid failing the CI.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,8 +1,11 @@
 name: Test kube-linter
 
 on:
-  - pull_request
-  - push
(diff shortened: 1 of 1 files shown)


viswajithiii (approved):

janisz:
I thought this line will handle that https://github.com/dhaus67/kube-linter/blob/c67ec801962acfbed048d1be93038a493061a9cb/.github/workflows/build.yaml#L123
Now we can delete it

PR #310, opened 2022-04-29
Added target-port template and check for it

### LINKED ISSUE #124: [FEATURE_REQUEST] Add check for valid Port Names / Length
Kube-Linter should check for valid port names.

Kubectl checks for port names and displays an error if they are to long.

```
Error from server (Invalid): Service "foobar-event-management" is invalid: spec.ports\[0\].targetPort: Invalid value: "foobar-event-management": must be no more than 15 characters
```

Could this check be also added to kube-linter somehow?

mtodor:
Fixes #124 

This PR adds a template and check for Service ports `targetPort` and Deployment ports `name` to check if they are valid. I have kept the scope small and addressed only this topic in this PR.


pkg/templates/targetport/template.go
```diff
@@ -0,0 +1,102 @@
...
+		}),
+	})
+}
+
+func findPodPorts(podSpec *customtypes.PodSpec) []diagnostic.Diagnostic {
+	var results []diagnostic.Diagnostic
+
+	containers := podSpec.AllContainers()
+	for _, container := range containers {
+		for _, port := range container.Ports {
```
janisz, line 62: I think we should skip if `port.Name` is empty as we have no way to detect if it was empty or null.

mtodor: @janisz, I have pushed the changes. We have a similar problem with `Service` if `targetPort` is not provided. It's the same as `int 0`. I have adjusted that too. I have also checked the outcome of the dry-run, and `targetPort: 0` is valid. 😞

docs/generated/checks.md
```diff
@@ -482,6 +482,21 @@ dirs:
 port: 22
 protocol: TCP
 ```
+## target-port-specification
```
viswajithiii, line 485: I would name this something like `invalid-target-ports`

pkg/templates/targetport/internal/params/params.go
```diff
@@ -0,0 +1,7 @@
+package params
+
+// Params represents the params accepted by this template.
+type Params struct {
+	// Set to true to validate targetPort names
+	TargetPort bool
```
viswajithiii, line 6: I don't think we need this param. This seems like a param that controls whether the check is enabled or not, which is not required, since enabling/disabling the check is controlled at a higher level. This check can be param-less.

viswajithiii (commented):
Looks pretty good, couple of tiny changes and it should be good to go

docs/generated/checks.md
mtodor: ✅

pkg/templates/targetport/internal/params/params.go
mtodor: ✅

mtodor:
@viswajithiii Tnx for the review! I have pushed the changes.

viswajithiii (approved):
Thanks @mtodor!

janisz (approved):

PR #318, opened 2022-05-05
Be more clear with errors found within anti-affinity rules.

### LINKED ISSUE #286: [BUG] no-anti-affinity rules does not work as expected
**System info:**
`stackrox/kube-linter:0.2.6`

**Describe the bug**
Even with anti-affinity rules apply to the pod, the `no-anti-affinity` rule fails

**To Reproduce**
```shell
# [Optional] Generate yaml manifest
helm repo add bitnami https://charts.bitnami.com/bitnami
helm template nginx bitnami/nginx --set replicaCount=2 --set containerSecurityContext.enabled=true --set containerSecurityContext.readOnlyRootFilesystem=true --set resources.requests.cpu=100m --set resources.limits.cpu=100m --set resources.requests.memory=100Mi  --set resources.limits.memory=100Mi > temp/temp.yaml

# Run kube linter
docker run --rm -v ${PWD}/temp:/data stackrox/kube-linter:0.2.6 lint /data/temp.yaml
```

**Sample YAML input**
```yaml
# Source: nginx/templates/deployment.yaml (stripped back)
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  replicas: 2
  selector:
    matchLabels:
      app.kubernetes.io/name: nginx
      app.kubernetes.io/instance: nginx
  template:
    metadata:
      labels:
        app.kubernetes.io/name: nginx
        app.kubernetes.io/instance: nginx
    spec:
      affinity:
        podAntiAffinity:
            - podAffinityTerm:
                labelSelector:
                  matchLabels:
                    app.kubernetes.io/name: nginx
                    app.kubernetes.io/instance: nginx
                namespaces:
                  - "default"
              weight: 1

      containers:
        - name: nginx
          image: docker.io/bitnami/nginx:1.21.6-debian-10-r0
          securityContext:
            readOnlyRootFilesystem: true
            runAsNonRoot: true
            runAsUser: 1001
          resources:
            limits:
              cpu: 100m
              memory: 100Mi
            requests:
              cpu: 100m
              memory: 100Mi

```

**Expected behavior**
All rules should pass

**Screenshots**
```shell
KubeLinter 0.2.6
/data/temp.yaml: (object: <no namespace>/nginx apps/v1, Kind=Deployment) object has 2 replicas but does not specify inter pod anti-affinity (check: no-anti-affinity, remediation: Specify anti-affinity in your pod specification to ensure that the orchestrator attempts to schedule replicas on different nodes. Using podAntiAffinity, specify a labelSelector that matches pods for the deployment, and set the topologyKey to kubernetes.io/hostname. Refer to https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity for details.)
```

**Additional context**
When `replicas` is set to 1, the rule correctly passes

dhaus67:
Fixes #286 .

Now, more specific errors will be returned when running checks against anti-affinity rules that have misconfigurations:

- Pod's and pod affinity's namespaces are not matching: `pod's namespace "pod-namespace" not found in anti-affinity's namespaces [affinity-namespace]`
- Topology key does not match: `anti-affinity's topology key does not match "some.key/value"`
- Pod's and pod affinity's labels are not matching: `pod's labels \"pod=labels\" do not match with anti-affinity's labels \"anti-affinity=labels\"`

### DIFF
--- a/pkg/templates/antiaffinity/template.go
+++ b/pkg/templates/antiaffinity/template.go
@@ -61,26 +61,30 @@ func init() {
 				if !hasPods {
 					return nil
 				}
+				affinity := podTemplateSpec.Spec.Affinity
+				// Short-circuit if no affinity rule is specified within the pod spec.
+				if affinity == nil || affinity.PodAntiAffinity == nil {
+					return []diagnostic.Diagnostic{
+						{Message: fmt.Sprintf("object has %d %s but does not specify inter pod anti-affinity",
+							replicas, stringutils.Ternary(replicas > 1, "replicas", "replica"))},
+					}
+				}
 				var foundIssues []diagnostic.Diagnostic
-				if affinity := podTemplateSpec.Spec.Affinity; affinity != nil && affinity.PodAntiAffinity != nil {
-					preferredAffinity := affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
-					requiredAffinity := affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution
-					for _, preferred := range preferredAffinity {
-						err := affinityTermMatchesLabelsAgainstNodes(preferred.PodAffinityTerm,
-							podTemplateSpec.Namespace, podTemplateSpec.Labels, topologyKeyMatcher)
-						if err == nil {
-							return nil
-						}
+				preferredAffinity := affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
+				requiredAffinity := affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+				for _, preferred := range preferredAffinity {
+					err := affinityTermMatchesLabelsAgainstNodes(preferred.PodAffinityTerm,
+						podTemplateSpec.Namespace, podTemplateSpec.Labels, topologyKeyMatcher)
+					if err != nil {
 						foundIssues = append(foundIssues, diagnostic.Diagnostic{
 							Message: err.Error(),
 						})
 					}
-					for _, required := range requiredAffinity {
-						err := affinityTermMatchesLabelsAgainstNodes(required, podTemplateSpec.Namespace,
-							podTemplateSpec.Labels, topologyKeyMatcher)
-						if err == nil {
-							return nil
-						}
+				}
+				for _, required := range requiredAffinity {
+					err := affinityTermMatchesLabelsAgainstNodes(required, podTemplateSpec.Namespace,
+						podTemplateSpec.Labels, topologyKeyMatcher)
+					if err != nil {
 						foundIssues = append(foundIssues, diagnostic.Diagnostic{
 							Message: err.Error(),
 						})
@@ -89,10 +93,7 @@ func init() {
 				if foundIssues != nil {
 					return foundIssues
 				}
-				return []diagnostic.Diagnostic{
-					{Message: fmt.Sprintf("object has %d %s but does not specify inter pod anti-affinity",
-						replicas, stringutils.Ternary(replicas > 1, "replicas", "replica"))},
-				}
+				return nil
 			}, nil
 		}),
 	})
--- a/pkg/templates/antiaffinity/template.go
+++ b/pkg/templates/antiaffinity/template.go
@@ -2,6 +2,7 @@ package antiaffinity
 
 import (
 	"fmt"
+	"strings"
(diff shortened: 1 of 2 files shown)


pkg/templates/antiaffinity/template.go
```diff
@@ -28,9 +29,10 @@ func defaultTopologyKeyMatcher(topologyKey string) bool {
 
 func init() {
 	templates.Register(check.Template{
-		HumanName:   "Anti affinity not specified",
-		Key:         templateKey,
-		Description: "Flag objects with multiple replicas but inter-pod anti affinity not specified in the pod template spec",
+		HumanName: "Anti affinity not specified",
+		Key:       templateKey,
+		Description: "Flag objects with multiple replicas but inter-pod anti affinity not specified in the pod " +
+			"template spec",
```
janisz, line 35: Why this change?

pkg/templates/antiaffinity/template.go
```diff
@@ -59,30 +61,47 @@ func init() {
...
 						}
+						foundIssues = append(foundIssues, diagnostic.Diagnostic{
+							Message: err.Error(),
+						})
 					}
 					for _, required := range requiredAffinity {
-						if affinityTermMatchesLabelsAgainstNodes(required, podTemplateSpec.Namespace, podTemplateSpec.Labels, topologyKeyMatcher) {
+						err := affinityTermMatchesLabelsAgainstNodes(required, podTemplateSpec.Namespace,
+							podTemplateSpec.Labels, topologyKeyMatcher)
+						if err == nil {
```
janisz, line 81: Should we do a short circuit here? What if we have an error in next iteration?

pkg/templates/antiaffinity/template.go
```diff
@@ -59,30 +61,47 @@ func init() {
...
 							return nil
 						}
+						foundIssues = append(foundIssues, diagnostic.Diagnostic{
+							Message: err.Error(),
+						})
 					}
 				}
+				if foundIssues != nil {
```
janisz, line 89: I think this should be in above `if` body

pkg/templates/antiaffinity/template.go
```diff
@@ -59,30 +61,47 @@ func init() {
 				if !hasPods {
 					return nil
 				}
+				var foundIssues []diagnostic.Diagnostic
 				if affinity := podTemplateSpec.Spec.Affinity; affinity != nil && affinity.PodAntiAffinity != nil {
```
janisz, line 65: How about inverting this condition?

janisz (changes requested):
I left some comments

pkg/templates/antiaffinity/template.go
dhaus67: Refactored to not exceed line-length of 130 characters.

pkg/templates/antiaffinity/template.go
dhaus67: I was wondering the same thing, I've kept the logic as before, short-circuiting on the first found, valid anti-affinity rule.
I'm fine with changing it to _not_ short-circuit, if we are fine with it.

pkg/templates/antiaffinity/template.go
janisz: Now this condition is redundant

pkg/templates/antiaffinity/template.go
```diff
@@ -59,30 +61,48 @@ func init() {
...
 				}
+				return nil
 			}, nil
 		}),
 	})
 }
 
-func affinityTermMatchesLabelsAgainstNodes(affinityTerm coreV1.PodAffinityTerm, podNamespace string, podLabels map[string]string, topologyKeyMatcher func(string) bool) bool {
-	// If namespaces is not specified in the affinity term, that means the affinity term implicitly applies to the pod's namespace.
+func affinityTermMatchesLabelsAgainstNodes(affinityTerm coreV1.PodAffinityTerm, podNamespace string,
```
viswajithiii, line 102: Nit: this function should be renamed now that it's no longer returning a `bool`. Perhaps `validateAffinityTermMatchesAgainstNodes`?

viswajithiii (approved):

pkg/templates/antiaffinity/template.go
dhaus67: Screenshot 2022-05-05 at 18 06 23

PR #325, opened 2022-05-06
Adding support for ServiceAccount

trumant:
Demonstrating support with additional tests focused
on using the forbidden-annotation check with objects
of Kind ServiceAccount

### DIFF
--- a/e2etests/testdata/forbidden-annotation-config.yaml
+++ b/e2etests/testdata/forbidden-annotation-config.yaml
@@ -1,13 +1,24 @@
 checks:
   addAllBuiltIn: false
 customChecks:
-  name: "forbid-annotation-reloader-stakater-auto"
-  description: ""
-  remediation: "Remove reloader.stakater.com/auto annotation"
-  scope:
-    objectKinds:
-      - DeploymentLike
-  template: "forbidden-annotation"
-  params:
-    key: "reloader.stakater.com/auto"
-    value: 'true'
+  - name: "forbid-annotation-reloader-stakater-auto"
+    description: ""
+    remediation: "Remove reloader.stakater.com/auto annotation"
+    scope:
+      objectKinds:
+        - DeploymentLike
+    template: "forbidden-annotation"
+    params:
+      key: "reloader.stakater.com/auto"
+      value: 'true'
+  - name: invalid-irsa-role
+    description: "IRSA annotations must have a valid IAM Role ARN value"
+    remediation: "Validate the format of the annotation's value to ensure it is a valid IAM Role ARN"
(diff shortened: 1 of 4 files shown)


e2etests/testdata/forbidden-annotation-config.yaml
```diff
@@ -1,13 +1,24 @@
...
+    description: "IRSA annotations must have a valid IAM Role ARN value"
+    remediation: "Validate the format of the annotation's value to ensure it is a valid IAM Role ARN"
+    scope:
+      objectKinds:
+        - ServiceAccount
+    template: "forbidden-annotation"
+    params:
+      key: "eks.amazonaws.com/role-arn"
+      value: |
+        !arn:aws:iam::\d{12}:role\/[a-zA-Z_+=,.@-]+$
```
janisz, line 24: Where this regexp come from?

trumant: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#:~:text=%3A%20No%20interruption-,RoleName,-A%20name%20for

janisz: > [RoleName is] a string of characters consisting of upper and lowercase alphanumeric characters with no spaces. You can also include any of the following characters: _+=,.@-.

Your regexp does not allow numeric values

I think role name part should be `[\w+=,.@-]{1,64}` as defined in [CreateRole](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html)

trumant: Awesome, thanks for the feedback and reference. I'll make adjustments

pkg/objectkinds/serviceaccount.go
```diff
@@ -0,0 +1,21 @@
+package objectkinds
```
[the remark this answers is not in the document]
trumant: Seems like there was a functional difference as when I removed it in [81779e5](https://github.com/stackrox/kube-linter/pull/325/commits/81779e56eef94c28092d18f880707eba3d9a8a09) we observed the bats test failure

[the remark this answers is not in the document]
viswajithiii: Ah yes, it needs to be there since you're creating a check just for service accounts. My bad, I missed that before.

trumant:
I think I've addressed all feedback requested. I appreciate your reviews when you have the time available @janisz

janisz:
@viswajithiii PTAL

PR #332, opened 2022-05-12
Remove circleci

janisz:
Since we moved all jobs to gh action there is no need to keep circleci.


dhaus67 (approved):

dhaus67 (commented):
Why is `ci/build` still run?

PR #353, opened 2022-06-02
Support check.ignorePaths config option

### LINKED ISSUE #127: [FEATURE_REQUEST] Add `--exclude-paths` option to exclude linting files under those paths
Our project contains a `third_party` folder containing yaml files that do not concern us, I would like an option to ignore all files under that path.

Currently this is not an option

**Additional context**
N\A

jumbosushi:
## Problem

We would like to exclude some directories from being checked while using the tool (fix #127)

## Solution

- Add `check.ignorePaths` configuration option that supports relative, homedir, and glob paths (didn't go with `excludePaths` since it might be confusing with the existing `exclude` option)

~This is a WIP~. Let me know if you would like to see more tests or have suggestions!

### DIFF
--- a/docs/configuring-kubelinter.md
+++ b/docs/configuring-kubelinter.md
@@ -48,7 +48,7 @@ checks:
 
 ## Ignore paths
 
-To ignore checks on files or direcotry under certain paths, add ignored paths to `ignorePaths`.
+To ignore checks on files or directories under certain paths, add ignored paths to `ignorePaths`.
 ```yaml
 checks:
   ignorePaths:
--- a/docs/configuring-kubelinter.md
+++ b/docs/configuring-kubelinter.md
@@ -46,6 +46,15 @@ checks:
   addAllBuiltIn: true
 ```
 
+## Ignore paths
+
+To ignore checks on files or direcotry under certain paths, add ignored paths to `ignorePaths`.
+```yaml
+checks:
+  ignorePaths:
+    - ~/foo/bar
+```
+
 > Equivalent CLI flag is `--add-all-built-in`
 
 > [!NOTE]
--- a/docs/configuring-kubelinter.md
+++ b/docs/configuring-kubelinter.md
@@ -53,6 +53,8 @@ To ignore checks on files or direcotry under certain paths, add ignored paths to
 checks:
   ignorePaths:
     - ~/foo/bar
+    - ../baz
+    - /tmp/*.yaml
 ```
 
 > Equivalent CLI flag is `--add-all-built-in`
--- a/pkg/configresolver/config_resolver.go
+++ b/pkg/configresolver/config_resolver.go
@@ -64,25 +64,25 @@ func GetIgnorePaths(cfg *config.Config) ([]string, error) {
 	errorList := errorhelpers.NewErrorList("check ignore paths")
 	ignorePaths := set.NewStringSet()
 	for _, path := range cfg.Checks.IgnorePaths {
-		switch {
-		case path[0] == '~':
-			expandedPath, err := homedir.Expand(path)
+		if path[0] == '~' {
+			path, err := homedir.Expand(path)
 			if err != nil {
 				errorList.AddStringf("could not expand path: %q", path)
 				continue
 			}
-			ignorePaths.Add(expandedPath)
-		case !filepath.IsAbs(path):
-			absolutePath, err := filepath.Abs(path)
+		}
+
+		if !filepath.IsAbs(path) {
+			path, err := filepath.Abs(path)
 			if err != nil {
 				errorList.AddStringf("could not expand non-absolute path: %q", path)
 				continue
 			}
-			ignorePaths.Add(absolutePath)
-		default:
-			ignorePaths.Add(path)
 		}
+
+		ignorePaths.AddAll(path)
 	}
+
 	if err := errorList.ToError(); err != nil {
 		return nil, err
 	}
--- a/pkg/configresolver/config_resolver.go
+++ b/pkg/configresolver/config_resolver.go
@@ -64,23 +64,24 @@ func GetIgnorePaths(cfg *config.Config) ([]string, error) {
 	errorList := errorhelpers.NewErrorList("check ignore paths")
 	ignorePaths := set.NewStringSet()
 	for _, path := range cfg.Checks.IgnorePaths {
+		result := path
 		if path[0] == '~' {
-			path, err := homedir.Expand(path)
+			expandedPath, err := homedir.Expand(path)
 			if err != nil {
-				errorList.AddStringf("could not expand path: %q", path)
+				errorList.AddStringf("could not expand path: %q", expandedPath)
 				continue
 			}
-		}
-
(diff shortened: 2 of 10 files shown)


pkg/lintcontext/create_contexts.go
```diff
@@ -50,6 +51,12 @@ func CreateContextsWithOptions(options Options, filesOrDirs ...string) ([]LintCo
 			continue
 		}
 
+        for _, path := range ignorePaths {
+            if strings.HasPrefix(fileOrDir, path) {
+                continue fileOrDirsLoop
+            }
+        }
```
jumbosushi, line 58: This is a naive solution to do prefix check by iterating `ignorePaths` slice. Let me know if you have suggestions for more optimized solution 🤔

docs/configuring-kubelinter.md
```diff
@@ -46,6 +46,15 @@ checks:
   addAllBuiltIn: true
 ```
 
+## Ignore paths
+
+To ignore checks on files or direcotry under certain paths, add ignored paths to `ignorePaths`.
```
janisz, line 51: Do we support only absolute paths?

janisz: How about supporting globs?

jumbosushi: Relative path is supported as well, and added glob path support. Updated doc in [03c80d1](https://github.com/stackrox/kube-linter/pull/353/commits/03c80d17f4c666035f37ec3d2e6a54fc4bb8fe7d)

pkg/configresolver/config_resolver_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"testing"
+
+	"github.com/mitchellh/go-homedir"
+	"github.com/stretchr/testify/assert"
+	"golang.stackrox.io/kube-linter/pkg/config"
+)
+
+var cfg *config.Config = new(config.Config)
+
+func TestHomeIgnorePath(t *testing.T) {
```
janisz, line 15: How about using table test for this tests?

pkg/configresolver/config_resolver.go
```diff
@@ -55,3 +58,36 @@ func GetEnabledChecksAndValidate(cfg *config.Config, checkRegistry checkregistry
...
+				continue
+			}
+			result = expandedPath
+		} else if !filepath.IsAbs(path) {
+			absPath, err := filepath.Abs(path)
+			if err != nil {
+				errorList.AddStringf("could not expand non-absolute path: %q", absPath)
+				continue
+			}
+			result = absPath
```
janisz, line 81: How about extracting inner body to a func?

```
func getIgnorePath(path string) (string, error) {
	switch {
	case path[0] == '~':
		expandedPath, err := homedir.Expand(path)
		if err != nil {
			return "", errors.Wrapf(err, "could not expand path: %q", expandedPath)
		}
		return expandedPath, nil
	case !filepath.IsAbs(path):
		absPath, err := filepath.Abs(path)
		if err != nil {
			return "", errors.Wrapf(err, "could not expand non-absolute path: %q", absPath)
		}
		return absPath, nil
	default:
		return path, nil
	}
}
```

janisz (commented):
Is it still a WIP?

pkg/configresolver/config_resolver_test.go
jumbosushi: good idea [90986c0](https://github.com/stackrox/kube-linter/pull/353/commits/90986c0d61305c473b671a7e9d26c7af7251503b)

docs/configuring-kubelinter.md
```diff
@@ -46,6 +46,17 @@ checks:
...
```
trumant, line 51:
```suggestion
To ignore checks on files or directories under certain paths, add ignored paths to `ignorePaths`.
```

pkg/lintcontext/create_contexts.go
```diff
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/bmatcuk/doublestar/v3"
```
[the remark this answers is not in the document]
jumbosushi: Updated to v4 in [9b3adf1](https://github.com/stackrox/kube-linter/pull/353/commits/9b3adf114a36ecc6be49a0ef8e9e6de0d00ce621). I saw v4 had [Go 1.16+ req](https://github.com/bmatcuk/doublestar/blob/5260d07/README.md#about) and I originally misread kube-linter's go.mod's go version

jumbosushi:
just following up on this @janisz @viswajithiii 👀

janisz:
@dhaus67 PTAL

dhaus67 (changes requested):
I've tested your changes with a sample configuration as such:
```yaml
checks:
  # if doNotAutoAddDefaults is true, default checks are not automatically added.
  doNotAutoAddDefaults: false

  # addAllBuiltIn, if set, adds all built-in checks. This allows users to
  # explicitly opt-out of checks that are not relevant using Exclude.
  # Takes precedence over doNotAutoAddDefaults, if both are set.
  addAllBuiltIn: false

  # include explicitly adds checks, by name. You can reference any of the built-in checks.
  # Note that customChecks defined above are included automatically.
  include:
  - "required-label-owner"
  # exclude explicitly excludes checks, by name. exclude has the highest priority: if a check is
  # in exclude, then it is not considered, even if it is in include as well.
  exclude:
  - "privileged"
  
  ignorePaths:
  - tests/**
```

and simply ran kube-linter in the repo-root like the following:
```
kube-linter --config config.yaml lint .
```

Now I'd expect, looking at the `ignorePaths` configuration, that the `tests` folder and all its subfolders would be ignored and skipped. Essentially, I would expect a behavior similar to `.gitignore`.

However, it seems as if we are only checking whether the path is a ignore path **before** [walking the file path](https://github.com/stackrox/kube-linter/pull/353/files#diff-1c457bc54ac821de3852c0e8970c4f9ace486a385f51493586cd2935a951996aR64) and will miss removing potentially ignored directory paths afterwards.

I feel like this could lead to difficulties using the configuration as-is, as people may expect a different behavior.

What would be best is to achieve the following:
- Check whether the path itself given for linting falls under any `ignoredPath`.
- When walking the file path, check each path whether it falls under any `ignoredPath`. If that's the case, ignore the path.

janisz:
@jumbosushi What's the status of this?

janisz:
@jumbosushi What's the status of this?

janisz (changes requested):
Please resolve conflicts

dhaus67:
@jumbosushi any interest in finalizing this PR?

If not, I'll take this one over after some time, since I feel like this is a good configuration to have.

jumbosushi:
Feel free to finish the feature! Not sure if i'll have time to complete it

titanlien:
Hi @dhaus67 Did you have any update for this PR?

titanlien:
I rebased some conflicts in go.mod then created a new PR there.
https://github.com/stackrox/kube-linter/pull/460

janisz:
Closing as #460 was merged

PR #373, opened 2022-07-08
Skip failing anti-affinity rules if at least one valid rule is found.

### LINKED ISSUE #369: [BUG] Lint fails on no-anti-affinity check in 0.3.0 where it succeeded in 0.2.6
**System info:**
 - OS: Linux

**Describe the bug**
`/dir/lintme.yaml: (object: linkerd/linkerd-destination apps/v1, Kind=Deployment) anti-affinity's topology key does not match "failure-domain.beta.kubernetes.io/zone" (check: no-anti-affinity, remediation: Specify anti-affinity in your pod specification to ensure that the orchestrator attempts to schedule replicas on different nodes. Using podAntiAffinity, specify a labelSelector that matches pods for the deployment, and set the topologyKey to kubernetes.io/hostname. Refer to https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity for details.)`

The pod being checked here has the below affinity rules:

```
      affinity:
        podAntiAffinity:
          - podAffinityTerm:
              labelSelector:
                matchExpressions:
                - key: component
                  operator: In
                  values:
                  - web
              topologyKey: failure-domain.beta.kubernetes.io/zone
            weight: 100
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: component
                operator: In
                values:
                - web
```
As you can see, we have 2 different topologyKey values in use here. This error started happening after we upgraded kube-linter to v0.3.0. I did some searching and I found #318 is likely the cause here.

I'm not sure why the output tells me to set the topologyKey to `kubernetes.io/hostname` when the failure also clearly states `anti-affinity's topology key does not match "failure-domain.beta.kubernetes.io/zone"` -- The remediation text should match the failure, but it doesn't, so it's not clear here which one of the anti affinity rules is the problem, or what the proper solution is.

**To Reproduce**
Set affinity rules as above on a deployment manifest and then lint it with kube-linter.

**Sample YAML input**
See above

**Expected behavior**
Lint successfully passes like it did in 0.2.6

dhaus67:
# Description

When adding [more verbose errors to no-anti-affinity check](https://github.com/stackrox/kube-linter/pull/318), we remove the short-circuiting on the first valid anti-affinity rule.
However, we should short-circuit on the first anti-affinity rule, as this is the expected behavior of the check.

Fixes #369

### DIFF
--- a/pkg/templates/antiaffinity/template.go
+++ b/pkg/templates/antiaffinity/template.go
@@ -72,6 +72,15 @@ func init() {
 				var foundIssues []diagnostic.Diagnostic
 				preferredAffinity := affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
 				requiredAffinity := affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+				// Short-circuit if affinity rule is specified but both preferred and required are empty.
+				if len(preferredAffinity) == 0 && len(requiredAffinity) == 0 {
+					return []diagnostic.Diagnostic{
+						{Message: fmt.Sprintf("object has %d %s but does not specify preferred or required "+
+							"inter pod anti-affinity during scheduling",
+							replicas, stringutils.Ternary(replicas > 1, "replicas", "replica"))},
+					}
+				}
+
 				for _, preferred := range preferredAffinity {
 					err := validateAffinityTermMatchesAgainstNodes(preferred.PodAffinityTerm,
 						podTemplateSpec.Namespace, podTemplateSpec.Labels, topologyKeyMatcher)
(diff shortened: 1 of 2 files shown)


pkg/templates/antiaffinity/template.go
```diff
@@ -75,20 +75,22 @@ func init() {
...
-							Message: err.Error(),
-						})
+					if err == nil {
+						return nil
 					}
+					foundIssues = append(foundIssues, diagnostic.Diagnostic{
+						Message: err.Error(),
+					})
 				}
 				return foundIssues
```
viswajithiii, line 95: This will return `nil` if no issues are found, which is wrong. (Which also suggests we need some unit tests here 😅)

dhaus67: Hm, but what would the expected behavior be if we didn't find any issues?

We are already short-circuiting if no anti-affinity rule is created beforehand.

viswajithiii: Ah, just saw that short-circuit. Well, we need to enhance it so that it also handles the case where podAntiAffinity is not `nil`, but both the arrays ("required" and "preferred") are empty. If we short-circuit in that case as well, then we should be good.

viswajithiii (changes requested):
See comment

janisz (approved):

Issue #387, opened 2022-08-08
[FEATURE_REQUEST] Add check for dangling ingress 


charlesoconor:
There are checks for dangling services and network policy's. It would be great to have one for ingresses as well. Just ran into an issue in my setup where someone forgot to create a k8s service the ingress was referencing and it would be awesome to catch this lint time. Happy to add the functionality.

PR #388, opened 2022-08-09
Create dangling ingress check 

charlesoconor:
Creating a check for dangling ingresses. I think it will fit in well with the other rules that already exist in this tool. Have been seeing these pop up from time to time in the projects I've been working on and would love to be able to catch them at lint time. 

I punted on dealing with resources that ingress can point to an only check for services, but could add that logic in if need be. My assumption was this would cover almost all of the use cases. 

Closes https://github.com/stackrox/kube-linter/issues/387


janisz:
Could you add e2e test?

charlesoconor:
Yes we can. Adding a e2e test

pkg/templates/danglingingress/template.go
```diff
@@ -0,0 +1,112 @@
...
+					service, ok := obj.K8sObject.(*v1.Service)
+					if !ok {
+						continue
+					}
+
+					serviceName := service.ObjectMeta.Name
+					if _, ok := selectors[serviceName]; ok {
+						delete(selectors, serviceName)
+						if len(selectors) == 0 {
+							// Found the all!
```
dhaus67, line 94: super nit:
```suggestion
							// Found them all!
```

pkg/templates/danglingingress/template.go
```diff
@@ -0,0 +1,112 @@
...
+	service := b.Service
+	if service == nil {
+		return ""
+	}
+
+	return service.Name
+}
+
+func getSelectorsFromIngress(ingress *networkingV1.Ingress) map[string]bool {
+	selectors := map[string]bool{}
```
dhaus67, line 31: I'd favor `map[string]struct{}` and setting `selectors[s] = struct{}{}` subsequently:
```suggestion
	selectors := map[string]struct{}
```

dhaus67 (commented):
Thanks for the contrib + adding the E2E tests so fast!

I would have one question regarding the definition of dangling ingress resource:
Currently, we define the ingress rule as dangling if the service name does not match. 

However, for the ingress to work correctly, the service's port name / number must match as well. This can easily be misconfigured.

I saw this multiple times in the past be the culprit of the ingress not working correctly, so I do see the opportunity here to improve this as well with the new check during linting.

I'd suggest we add this to the dangling check as well. Then, we will check that both the service name match as well as defined port (either via its name / number).

pkg/lintcontext/mocks/ingress.go
```diff
@@ -0,0 +1,29 @@
...
+			Kind:       objectkinds.Ingress,
+			APIVersion: objectkinds.GetIngressAPIVersion(),
+		},
+		ObjectMeta: metaV1.ObjectMeta{Name: name},
+		Spec:       networkingV1.IngressSpec{},
+	}
+}
+
+// ModifyIngess modifies a given networkpolicy in the context via the passed function.
+func (l *MockLintContext) ModifyIngess(t *testing.T, name string, f func(ingress *networkingV1.Ingress)) {
```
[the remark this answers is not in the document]
charlesoconor: :+1:

pkg/templates/danglingingress/template.go
charlesoconor: :+1:

pkg/templates/danglingingress/template.go
charlesoconor: :+1:

pkg/templates/danglingingress/template.go
```diff
@@ -0,0 +1,112 @@
...
+	return selectors
+}
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Dangling Ingress",
+		Key:         templateKey,
+		Description: "Flag ingress which do not match any service",
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.Service},
```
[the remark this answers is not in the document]
charlesoconor: yes it should be

charlesoconor:
I think it's a great idea to check the ports as well. Update the PR for that.

docs/generated/checks.md
```diff
@@ -67,6 +67,15 @@ verbs:
 **Remediation**: Confirm that your HorizontalPodAutoscaler's scaleTargetRef correctly matches one of your deployments.
 
 **Template**: [dangling-horizontalpodautoscaler](generated/templates.md#dangling-horizontalpodautoscalers)
+## dangling-ingress
+
+**Enabled by default**: No
+
+**Description**: Indicates when ingress do not have any associated services.
+
+**Remediation**: Confirm that your ingress's backend correctly matches the name on one of your services.
```
dhaus67, line 76:
```suggestion
**Remediation**: Confirm that your ingress's backend correctly matches the name and port on one of your services.
```

pkg/builtinchecks/yamls/dangling-ingress.yaml
```diff
@@ -0,0 +1,7 @@
+name: "dangling-ingress"
+description: "Indicates when ingress do not have any associated services."
+remediation: "Confirm that your ingress's backend correctly matches the name on one of your services."
```
dhaus67, line 3:
```suggestion
remediation: "Confirm that your ingress's backend correctly matches the name and port on one of your services."
```

dhaus67 (approved):
Only two minor doc comments, otherwise LGTM!

Thanks for making the changes quickly!

docs/generated/checks.md
charlesoconor: :+1:

pkg/builtinchecks/yamls/dangling-ingress.yaml
charlesoconor: :+1:

charlesoconor:
Thanks for the approval! Excited to be able to use this check I've been running into issues around this a lot lately.

PR #393, opened 2022-08-10
fix: Update object kinds for lint templates

heckler1:
This expands the object kinds for a number of lint checks, to include resources that are accessed by the lint check function. I have gone through all of the templates and these are the only ones I found where the objects accessed in the function do not match the ObjectKinds.

This means that the ObjectKinds field more accurately reflects the reality of what resources a lint check expects to be able to access. Any unsupported objects are still filtered by the checks, so this does not effect the function of the checks.

I believe clarifying the intent of these templates is a good thing in-and-of itself, but there are also other usecases, especially when running kube-linter in-cluster.

I hope this is trivial enough, but happy to raise an issue or discuss further if that helps.

### DIFF
--- a/docs/generated/templates.md
+++ b/docs/generated/templates.md
@@ -108,7 +108,7 @@ KubeLinter supports the following templates:
 
 **Description**: Flag HorizontalPodAutoscalers that target a resource that does not exist
 
-**Supported Objects**: HorizontalPodAutoscaler
+**Supported Objects**: HorizontalPodAutoscaler,DeploymentLike
 
 
 ## Dangling NetworkPolicies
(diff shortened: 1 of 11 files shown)


janisz:
>  I have gone through all of the templates and these are the only ones I found where the objects accessed in the function do not match the ObjectKinds.

Do you think we can automate this check and add it as a part of CI?

heckler1:
I don't see a way to check in code which types are being accessed, but if I'm missing something I'm more than happy to add that support.

janisz (commented):
I think I don't understand why we need this change? Could you add tests that were failing before but now are passing?

janisz:
@heckler1 Any update on this?

heckler1:
@janisz sorry for the delay, I've been meaning to come back to this. I have been considering this change more and I don't really think it makes as much sense (both for this project as well as my use case) as I initially believed. Sorry for the noise here!

PR #395, opened 2022-08-22
feat: Add ephemeral containers to PodSpec.AllContainers()

heckler1:
EphemeralContainers were promoted to beta in 1.23, so we should add support for them.

This also includes a couple of small comment fixes from running `make lint` locally.


janisz (approved):

janisz:
Thanks!

PR #419, opened 2022-09-30
Bump golang.org/x/net

tspearconquest:
Patch CVE-2022-27664


janisz:
Thanks for PR. I think we should split it in two.
1. Bump version
2. gofmt generated code or edit template so it will generate formatted code.

tspearconquest:
I've reverted the go fmt related changes.

tspearconquest:
https://stackoverflow.com/questions/34126083/is-it-possible-to-use-gofmt-on-templates-that-are-designed-to-be-used-with-go-ge indicates it's not really sane to try to format templates, but that being said, the pkg/config/flags.go does need formatting so I've opened #421 to fix that.

janisz (approved):

janisz:
Thanks for this PR. I'd like to not that we are not affected by this CVE as kube-linter uses Go 1.19.1 https://github.com/stackrox/kube-linter/actions/runs/3149228054/jobs/5120616807#step:4:16 and made no network connections. net dependency is `indirect`. 

Nevertheless it's good to stay on top with latest updated.

PR #421, opened 2022-09-30
Go fmt

tspearconquest:


pkg/config/flags.go
```diff
@@ -1,4 +1,5 @@
 // Code generated by kube-linter flag codegen. DO NOT EDIT.
```
janisz, line 1: This is autogenerated file. Please edit generator https://github.com/stackrox/kube-linter/blob/ab8a80b4653c8f08fb7dd395aa1a0bc1b7d04bc0/pkg/templates/codegen/main.go#L40

Issue #455, opened 2022-11-25
[FEATURE_REQUEST] validate that probe port match a existing port


jouve:
**Description of the problem/feature request**

when defining probes (liveness, readiness, startup) , the port can be specified as either an int or a string.
In that case, the string must match the name of a port defined in ports.

I suggest a check to validate that when the probe port is a string, it matches one the port by name.

**Additional context**

in case of mismatch, the yaml is accepted by the kubernetes, but the probe will fail latter : `Liveness probe errored: strconv.Atoi: parsing "xxx": invalid syntax`

would be ok:
```
containers:
- ports:
  - containerPort: 1234
    name: probe
  livenessProbe:
    httpGet:
      port: probe
```

would be an error:

```
containers:
- ports:
  - containerPort: 1234
    name: probe
  livenessProbe:
    httpGet:
      port: foo
```

PR #457, opened 2022-11-26
Check that string probe port match container port

jouve:
implementation of #455

if a probe port is a string and the string is not of number (like `"1234"` ) and the string is not in the container ports (ie. ` ports[].name), then raise a diagnostic message.

while the resource it accepted by kube currently, the liveness/readiness/startup will fail anyway, so added to default checks.


pkg/templates/probeport/template.go
```diff
@@ -0,0 +1,59 @@
...
+					if httpGet := probe.HTTPGet; httpGet != nil {
+						port = httpGet.Port
+					} else if tcpSocket := probe.TCPSocket; tcpSocket != nil {
+						port = tcpSocket.Port
+					} else {
+						continue
+					}
+					if port.Type == intstr.String && !portNames.Contains(port.StrVal) && port.IntValue() == 0 {
+						results = append(results, diagnostic.Diagnostic{
+							Message: fmt.Sprintf("probe port %s does not a port in container %q.", port.StrVal, container.Name),
```
janisz, line 51: Something is wrong with this sentence. I think it misses a verb.
```suggestion
							Message: fmt.Sprintf("probe port %s does not match a port in container %q.", port.StrVal, container.Name),
```

pkg/templates/probeport/template.go
```diff
@@ -0,0 +1,59 @@
...
+		HumanName:   "Probe Port",
+		Key:         "probe-port",
+		Description: "Flag unknown probe port",
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.DeploymentLike},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
+			return util.PerNonInitContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
```
janisz, line 29: It will be great if you extract this to a named function and add unit tests to check all code paths.

janisz (changes requested):

janisz:
@jouve You need to regenerate files

jouve:
I'll update this PR soon :)

janisz:
@jouve ping

janisz:
@jouve Could you rebase? I think we can make it into next release

janisz:
Closing due to lack of activity
This probably could be implemented as a CEL expression:
- #1012

PR #461, opened 2022-12-13
Add check for duplicate name env var

charlesoconor:
I've got some complicated helm templates and have been running into issues where devs add the same variable in different helpers. This check would make sure we don't add the same one more than once. K8s will apply the last one added for you but this can be unintuitive.


pkg/templates/duplicatenvvar/template_test.go
```diff
@@ -0,0 +1,109 @@
...
+			Diagnostics: map[string][]diagnostic.Diagnostic{
+				targetName: {
+					// Ensure we only get one error for ENV_1
+					{Message: "Duplicate environment variable ENV_1 in container \"container\" found"},
+					{Message: "Duplicate environment variable ENV_2 in container \"container\" found"},
+				},
+			},
+		},
+	})
+}
```
janisz, line 109: How about using table driven test? I think assertions might work weird but we can fix it in a follow up. We need to pass proper `T()` inside validate.
```suggestion
func (s *DuplicateEnvVarTestSuite) TestDeploymentWith() {
	const targetName = "deployment01"
	testCases := []struct {
		name      string
		container v1.Container
		expected  map[string][]diagnostic.Diagnostic
	}{
		{
			name: "WithNoDuplicatesShouldPass",
			container: v1.Container{
				Env: []v1.EnvVar{
					{Name: "ENV_1"},
					{Name: "ENV_2"},
				},
			},
			expected: nil,
		},
		{
			name: "DuplicatesShouldReportErrors",
			container: v1.Container{
				Env: []v1.EnvVar{
					{Name: "ENV_1"},
					{Name: "ENV_1"},
				},
			},
			expected: map[string][]diagnostic.Diagnostic{
				targetName: {
					// Ensure we only get one error for ENV_1
					{Message: "Duplicate environment variable ENV_1 in container \"\" found"},
				},
			},
		},
		{
			name: "MultipleDuplicatesShouldReportThemAllButOnlyOncePerEnv",
			container: v1.Container{
				Name: "container",
				Env: []v1.EnvVar{
					{Name: "ENV_1"},
					{Name: "ENV_1"},
					{Name: "ENV_1"},
					{Name: "ENV_2"},
					{Name: "ENV_2"},
					{Name: "ENV_3"},
				},
			},
			expected: map[string][]diagnostic.Diagnostic{
				targetName: {
					// Ensure we only get one error for ENV_1
					{Message: "Duplicate environment variable ENV_1 in container \"container\" found"},
					{Message: "Duplicate environment variable ENV_2 in container \"container\" found"},
				},
			},
		},
	}
	for _, tc := range testCases {
		s.Run(tc.name, func() {
			s.ctx.AddMockDeployment(s.T(), targetName)
			s.ctx.AddContainerToDeployment(s.T(), targetName, tc.container)
			s.Validate(s.ctx, []templates.TestCase{{
				Param:       params.Params{},
				Diagnostics: tc.expected,
			}})
		})
	}
}
```

pkg/templates/duplicatenvvar/template_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	ctx *mocks.MockLintContext
+}
+
+func (s *DuplicateEnvVarTestSuite) SetupTest() {
+	s.Init(templateKey)
+	s.ctx = mocks.NewMockContext()
+}
+
+func (s *DuplicateEnvVarTestSuite) TestDeploymentWithNoDuplicatsPass() {
```
janisz, line 29:
```suggestion
func (s *DuplicateEnvVarTestSuite) TestDeploymentWithNoDuplicatesPass() {
```

pkg/templates/duplicatenvvar/template.go
```diff
@@ -0,0 +1,51 @@
...
+						results = append(results, diagnostic.Diagnostic{
+							Message: fmt.Sprintf(
+								"Duplicate environment variable %s in container %q found",
+								envVar.Name,
+								container.Name,
+							),
+						})
+					}
+
+					envVarNames[envVar.Name]++
```
janisz, line 45: nit
```suggestion
					envVarNames[envVar.Name]++
					if num, ok := envVarNames[envVar.Name]; !ok || num != 2 {
						continue
					}
					results = append(results, diagnostic.Diagnostic{
						Message: fmt.Sprintf(
							"Duplicate environment variable %s in container %q found",
							envVar.Name,
							container.Name,
						),
					})
```

janisz (approved):
LGTM
Just one typo to fix and some nitpicks but if @dhaus67 approves we can merge it and fix them in a separated PR.

pkg/templates/duplicatenvvar/template_test.go
charlesoconor: Will convert this

dhaus67 (approved):

charlesoconor:
Thanks for the Approval! Can someone merge this for me? I don't have write access.

PR #465, opened 2022-12-14
Allow ignoring specific labels in dangling service selector check

heckler1:
We have services that select pods based on `statefulset.kubernetes.io/pod-name`, which is populated by the Kubernetes control plane. I would like to prevent this from being linted on for our use-case, while still checking if the other selector labels are correct.

This instance with service labels is specific to my use-case, but I'm also happy to expand this to the other "dangling" checks if desired.

### DIFF
--- a/pkg/templates/danglingservice/template.go
+++ b/pkg/templates/danglingservice/template.go
@@ -16,17 +16,19 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 )
 
+const templateKey = "dangling-service"
+
 func init() {
 	templates.Register(check.Template{
 		HumanName:   "Dangling Services",
-		Key:         "dangling-service",
+		Key:         templateKey,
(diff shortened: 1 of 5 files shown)


pkg/templates/danglingservice/template.go
```diff
@@ -42,6 +42,11 @@ func init() {
 						Message: "service has no selector specified",
 					}}
 				}
+
+				for _, ignoredLabel := range p.IgnoredLabels {
+					delete(selector, ignoredLabel)
+				}
+
```
janisz, line 49: Could you add a test for this change?

janisz (approved):

dhaus67 (approved):

heckler1:
@dhaus67, @janisz If I could trouble one of y'all for a merge, I'd appreciate it. Thanks!

PR #485, opened 2023-02-03
feat: add PodDisruptionBudget kind to registered objectkinds

tremes:


janisz (commented):
Could you add test for this change?

tremes:
@janisz yes I can. I am just wondering how these object kinds are tested......

janisz:
Probably the best way to add object kind is with a template that uses it. Then you can add it in mocks and use in unit tests and/or add a e2e tests.

Here is a tutorial how to add a new check https://www.psaggu.com/upstream-contribution/2021/08/17/notes.html

janisz:
Closing in favour of https://github.com/stackrox/kube-linter/pull/507

PR #497, opened 2023-02-21
fix: add `v` prefix to release-drafter's tag template

### LINKED ISSUE #494: kube-linter go module versioning is broken
### Problem statement

It seems that `kube-linter` is not adhering to semver versioning w.r.t having a `v` prefix in tag names, as mentioned e.g. in https://go.dev/blog/publishing-go-modules#semantic-versions-and-modules.

This causes `go` toolchain to not detect newer versions than `v0.4.0` which by accident (?) had `v` prefix.

### How to reproduce

To reproduce this it's enough to:

```console
$ go mod init randomname
$ go get golang.stackrox.io/kube-linter@latest
go: added golang.stackrox.io/kube-linter v0.4.0
```

---

One thing that might be considered is changing https://github.com/stackrox/kube-linter/blob/main/.github/workflows/build.yaml#L10 to `"v*"` or something more specific like:

```
      - 'v[0-9]+.[0-9]+.[0-9]+'
      - 'v[0-9]+.[0-9]+.[0-9]+-alpha.[0-9]+'
      - 'v[0-9]+.[0-9]+.[0-9]+-beta.[0-9]+'
      - 'v[0-9]+.[0-9]+.[0-9]+-rc.[0-9]+'
```

czeslavo:
In order to make tags created by the release-drafter compatible with Go modules, tags have to begin with `v` prefix.

It should fix #494.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,5 +1,5 @@
 name-template: "v$NEXT_PATCH_VERSION 🌈"
-tag-template: "$NEXT_PATCH_VERSION"
+tag-template: "v$NEXT_PATCH_VERSION"
 categories:
   - title: "🚀 Features"
     labels:


janisz:
I don't know what was the reason for dropping `v` from tag. But I think it should not break anything.

janisz (approved):

janisz:
Maybe we should manually double tag all previous releases with `v` prefixed tags.

dhaus67 (approved):
Looks good to me. One thing to note is that the version tag will now change for users of the kube-linter action going forward, but as long as we keep the previously tagged version, I don't see any issue with making the change 👍

PR #507, opened 2023-03-07
Create check to ensure PDB have MaxUnavailable and MinAvailable set effectively

fahlmant:
Creates two checks for PDBs:
- Recommends changes to PDBs that have MaxUnavailable set to 0
- Recommends changes to PDBs that have MinAvailable set to the Replica value from related Deployments

This is a WIP. Tests still need to be added


pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,97 @@
...
+					return []diagnostic.Diagnostic{{
+						Message: fmt.Sprintf("pdb has invalid podSelector: %v", err),
+					}}
+				}
+
+				// Evaluate deployments in the lintContext to see if they have MinAvailable set too low
+				for _, deployment := range getDeployments(lintCtx, pdb.Namespace, labelSelector) {
+					if *deployment.Spec.Replicas == pdb.Spec.MinAvailable.IntVal {
+						results = append(results, diagnostic.Diagnostic{
+							Message: fmt.Sprintf("Deployment %s has replicas equal to the minumum available replicas set by its PDB.", deployment.Name),
```
tremes, line 69: Typo in `minumum`

docs/generated/checks.md
```diff
@@ -410,6 +410,15 @@ strategyTypeRegex: ^(RollingUpdate|Rolling)$
 **Remediation**: Ensure pod does not accept unsafe traffic by isolating it with a NetworkPolicy. See https://cloud.redhat.com/blog/guide-to-kubernetes-ingress-network-policies for more details.
 
 **Template**: [non-isolated-pod](templates.md#non-isolated-pods)
+## pod-disruption-budget
+
+**Enabled by default**: No
+
+**Description**: Indicates when a DeploymentLike object do not have a "Pod Disruption Budget" specified provided it's replica count is more than the minimum specified value
```
tremes, line 417: > .... its replica count ?

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,97 @@
...
+
+				// Get the PDB provided
+				pdb, ok := object.K8sObject.(*pdbV1.PodDisruptionBudget)
+				if !ok {
+					return nil
+				}
+
+				// If MaxUnavailable is set to 0,
+				// return as no other checks will uniquely apply
+				if pdb.Spec.MaxUnavailable.IntVal == 0 {
```
tremes, line 48: I think this can fail with nil pointer when `MaxUnavailable` is not set.

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,97 @@
...
+	var deploymentList []v1.Deployment
+
+	for _, obj := range lintCtx.Objects() {
+		deployment, ok := obj.K8sObject.(*v1.Deployment)
+		if !ok {
+			continue
+		}
+
+		// Find any deployments in the same namespace with labels that match the PDBs selector
+		if labelSelector.Matches(labels.Set(deployment.Labels)) && namespace == deployment.Namespace {
```
tremes, line 91: Just question @fahlmant please. This requires the deployment to have the labels set, right? Wouldn't be better to check `labelSelector.Matches(labels.Set(deployment.Spec.Selector.MatchLabels))` ?

pkg/templates/pdb/template.go
fahlmant: Fixing

docs/generated/checks.md
fahlmant: Fixing

pkg/templates/pdb/template.go
fahlmant: Yes, I am adding the tests still and caught this, will fix the code to make sure it checks that.

pkg/templates/pdb/template.go
fahlmant: You're right, I was thinking about the pods the PDB checks, rather than the deployment backing the pods. Fixing

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,97 @@
...
+		}),
+	})
+}
+
+func getDeployments(lintCtx lintcontext.LintContext, namespace string, labelSelector labels.Selector) []v1.Deployment {
+
+	var deploymentList []v1.Deployment
+
+	for _, obj := range lintCtx.Objects() {
+		deployment, ok := obj.K8sObject.(*v1.Deployment)
```
tremes, line 85: And what about other controller objects? Would it make sense to check more? E.g `StatefulSet`, maybe `DeploymentConfig`, ...

fahlmant: The original proposal was just for Deployments, but I think a follow up PR to add this is a good idea.

docs/generated/checks.md
```diff
@@ -410,6 +410,15 @@ strategyTypeRegex: ^(RollingUpdate|Rolling)$
 **Remediation**: Ensure pod does not accept unsafe traffic by isolating it with a NetworkPolicy. See https://cloud.redhat.com/blog/guide-to-kubernetes-ingress-network-policies for more details.
 
 **Template**: [non-isolated-pod](templates.md#non-isolated-pods)
+## pod-disruption-budget
+
+**Enabled by default**: No
+
+**Description**: Indicates when a DeploymentLike object do not have a "Pod Disruption Budget" specified provided its replica count is more than the minimum specified value
```
dhaus67, line 417:
```suggestion
**Description**: Indicates when a DeploymentLike object does not have a "Pod Disruption Budget" specified provided its replica count is more than the minimum specified value
```

pkg/builtinchecks/yamls/pdbs.yaml
```diff
@@ -0,0 +1,8 @@
+name: "pod-disruption-budget"
+description: "Indicates when a DeploymentLike object do not have a \"Pod Disruption Budget\" specified provided its replica count is more than the minimum specified value"
+remediation: "Create a \"Pod Disruption Budget\" in the specified namespace with a \"MaxUnavailableValue\" set equal to or more than the specified criteria."
+scope:
+  objectKinds:
+    - PodDisruptionBudget
+    - DeploymentLike
+template: "pod-disruption-budget"
```
dhaus67, line 8: Please add a newline here.

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,114 @@
...
+					return nil
+				}
+
+				// If MaxUnavailable is set to 0,
+				// return as no other checks will uniquely apply
+				if pdb.Spec.MaxUnavailable.IntVal == 0 {
+					return []diagnostic.Diagnostic{{
+						Message: maxUnavailableZeroMsg,
+					},
+					}
```
dhaus67, line 52: nit in terms of formatting:
```suggestion
					return []diagnostic.Diagnostic{
					        {
						        Message: maxUnavailableZeroMsg,
					        },
					}
```

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,114 @@
...
+				}
+
+				labelSelector, err := metaV1.LabelSelectorAsSelector(&metaV1.LabelSelector{
+					MatchLabels:      pdb.Spec.Selector.MatchLabels,
+					MatchExpressions: pdb.Spec.Selector.MatchExpressions},
+				)
+
+				if err != nil {
+					return []diagnostic.Diagnostic{{
+						Message: fmt.Sprintf("pdb has invalid podSelector: %v", err),
```
dhaus67, line 62:
```suggestion
						Message: fmt.Sprintf("PDB has invalid label selector: %v", err),
```

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,114 @@
...
+		}),
+	})
+}
+
+func getDeployments(lintCtx lintcontext.LintContext, namespace string, labelSelector labels.Selector) ([]v1.Deployment, error) {
+
+	var deploymentList []v1.Deployment
+
+	for _, obj := range lintCtx.Objects() {
+		deployment, ok := obj.K8sObject.(*v1.Deployment)
```
dhaus67, line 93: Instead of doing this in a non-generic way that only covers `v1.Deployment`, please do the following:

Instead of returning a list of `[]v1.Deployment`, return a list of `[]k8sutil.Object`. You can achieve this by:
- removing the explicit cast to `v1.Deployment`.
- skip objects which have no replicas, i.e. `_, exists := extract.Replicas(obj.K8sObject); exists == false`
- use `extract.Selector(obj.K8sObject)` to retrieve the object's label selector.

Within `line75` where we compare the replicas, you can use `extract.Replicas(k8sObj)` instead and keep things generic to cover all "deployment like" objects.

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,114 @@
...
+					return []diagnostic.Diagnostic{{
+						Message: fmt.Sprintf("pdb has invalid podSelector: %v", err),
+					}}
+				}
+
+				// Evaluate deployments in the lintContext to see if they have MinAvailable set too low
+				deployments, err := getDeployments(lintCtx, pdb.Namespace, labelSelector)
+				if err != nil {
+					return []diagnostic.Diagnostic{{
+						Message: err.Error(),
```
dhaus67, line 70: nit: would be great to be more specific within this error message, i.e.:
```suggestion
						Message: fmt.Sprintf("Failed to retrieve deployments matching the PDB's label selector within namespace %s: %v", namespace, err.Error()),
```

pkg/templates/pdb/template.go
```diff
@@ -0,0 +1,114 @@
...
+				return results
+
+			}, nil
+		}),
+	})
+}
+
+func getDeployments(lintCtx lintcontext.LintContext, namespace string, labelSelector labels.Selector) ([]v1.Deployment, error) {
+
+	var deploymentList []v1.Deployment
```
dhaus67, line 90: nit (in conjunction with the comment below on the refactoring of the function):
```suggestion
	objectList := make([]k8sutil.Object, 0, len(lintCtx.Objects()))
```

pkg/templates/pdb/template.go
fahlmant: The original scope of the check was just for Deployments, widening the scope will delay the PR a bit as the tests will need to be more extensive.

pkg/templates/pdb/template.go
tremes: OK...I am fine with a follow-up PR. Thanks

tremes:
It's still not clear to me when and how this new check can be enabled. It seems it won't be enabled by default so will it be a custom check?

fahlmant:
@tremes the PR is still a WIP

fahlmant:
Tests currently are not working, adding a Deployment to the lintContext seems to not be working as the mocked object is showing that there's no Kind specified

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,178 @@
...
+				if !ok {
+					return nil
+				}
+
+				// Extract the MinAvailable value from the spec and check if it's a number or percentage
+				value, isPercent, err := getIntOrPercentValueSafelyFromString(pdb.Spec.MinAvailable.String())
+				if err != nil {
+					return []diagnostic.Diagnostic{
+						{
+							Message: fmt.Sprintf("pdb has invalid MinAvailable value: %v", err),
```
tremes, line 53: When I have a pdb, which doesn't set `minAvailable` then this leads to a failed check with message:
```
pdb has invalid MinAvailable value: invalid type: string \"<nil>\" is not a percentage
```
and remediation:
```
"Change the \"Pod Disruption Budget\" to have a \"MinAvailableValue\" set to number lower than the Replica count in the related DeploymentLike."
```
I guess it's not desired or am I wrong?

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,178 @@
...
+						},
+					}
+				}
+
+				for _, dl := range deploymentLikes {
+					pdbMinAvailable := value
+					replicas, _ := extract.Replicas(dl)
+					if isPercent {
+						// Calulate the actual value of the MinAvailable with respect to the Replica count if a percentage is set
+						pdbMinAvailable = int(replicas) * (value / 100)
```
tremes, line 99: Won't it be always zero? I guess you need something like  `pdbMinAvailable = float64(replicas) * (float64(value) / float64(100))`

fahlmant: In the end this still has to be an integer. It doesn't matter too much though because as long as it's less than 100%, it should be a valid config. For accuracy though, I think this will work:
`pdbMinAvailable = int(math.Floor(float64(replicas) * (float64(value) / float64(100))))`

This finds the exact value of minavailable based on a percentage, and then rounds down so cases like 1 replica with a 50% minavaiable (which calculates to 0.5) won't cause this remediation to fire. I'm not sure exactly how kubernetes handles this sort of case, so I'm erring on the side of permissive at first. It might be a good place for iteration

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+				if pdb.Spec.MinAvailable == nil {
+					return results
+				}
+
+				// Extract the MinAvailable value from the spec and check if it's a number or percentage
+				value, isPercent, err := getIntOrPercentValueSafelyFromString(pdb.Spec.MinAvailable.String())
+				if err != nil {
+					return []diagnostic.Diagnostic{
+						{
+							Message: fmt.Sprintf("pdb has invalid MinAvailable value: %v", err),
```
janisz, line 59: in other places PDB is capital. 
```suggestion
							Message: fmt.Sprintf("PDB has invalid MinAvailable value: %w", err),
```

janisz: also in other places we use `errors.Wrap`

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+		HumanName:   "Pod Disruption Budget Configuration Checks",
+		Key:         templateKey,
+		Description: "Flag non-optimal PDB configurations",
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{
+				objectkinds.PodDisruptionBudget},
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
```
janisz, line 38: since this function is quite long how about extracting it to a regular function (like `func name(....)`?

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+						{
+							Message: fmt.Sprintf("pdb has invalid MinAvailable value: %v", err),
+						},
+					}
+				}
+
+				// If the value is a percentage, handle the case where the MinValue is set to 100%
+				// as DeploymentLike's replica counts don't need to be compared
+				if isPercent {
+					if value == 100 {
```
janisz, line 67: nit: 
```suggestion
				if isPercent && value == 100 {
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+						},
+					}
+				}
+
+				for _, dl := range deploymentLikes {
+					pdbMinAvailable := value
+					replicas, _ := extract.Replicas(dl)
+					if isPercent {
+						// Calulate the actual value of the MinAvailable with respect to the Replica count if a percentage is set
+						pdbMinAvailable = int(math.Floor(float64(replicas) * (float64(value) / float64(100))))
```
janisz, line 105:
```suggestion
						pdbMinAvailable = int(replicas) * value / 100
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+
+		// Build Deployment labelSelector
+		// If there are no selectors on the object, then the PDB won't match the same pods as the Deployment Like
+		selectors, exists := extract.Selector(obj.K8sObject)
+		if exists {
+			// If there are no Replicas set on the Deployment Like, it's not possible to compare to a PDB
+			if _, exists := extract.Replicas(obj.K8sObject); exists {
+				objLabelSelector, err := metaV1.LabelSelectorAsSelector(selectors)
+				if err != nil {
+					return objectList, err
```
janisz, line 143:
```suggestion
					return nil, err
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+		}
+	}
+
+	return objectList, nil
+}
+
+func getIntOrPercentValueSafelyFromString(intOrStr string) (int, bool, error) {
+
+	var IntOrPercentValue = string("int")
```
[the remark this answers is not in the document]
janisz: How about using const for this. I think that strings are easy to make a mistake
```
	const (
		INT = iota
		PERCENT
	)
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,184 @@
...
+			return 0, false, fmt.Errorf("invalid type: string %q is not a percentage", intOrStr)
+		}
+		v, err := strconv.Atoi(s)
+		if err != nil {
+			return 0, false, fmt.Errorf("invalid value %q: %v", intOrStr, err)
+		}
+		return v, isPercent, nil
+	}
+	return 0, false, fmt.Errorf("invalid type: string %q neither int nor percentage", intOrStr)
+}
```
janisz, line 184: I think it could be simplified
```suggestion
func getIntOrPercentValueSafelyFromString(intOrStr string) (int, bool, error) {
	intValue, ifString := strconv.Atoi(intOrStr)
	if ifString == nil {
		return intValue, false, nil
	}

	s := intOrStr
	if !strings.HasSuffix(s, "%") {
		return 0, false, fmt.Errorf("invalid type: string %q is not a percentage", intOrStr)
	}
	s = strings.TrimSuffix(s, "%")
	v, err := strconv.Atoi(s)
	if err != nil {
		return 0, false, fmt.Errorf("invalid value %q: %v", intOrStr, err)
	}
	return v, true, nil
}
```

janisz (changes requested):

pkg/templates/pdbminavailable/template.go
fahlmant: I already had this and was asked to change it to use float64s :)

fahlmant: https://github.com/stackrox/kube-linter/pull/507#discussion_r1156852629

e2etests/bats-tests.sh
```diff
@@ -574,6 +574,20 @@ get_value_from() {
...
+  run ${cmd}  
+}
+
+@test "pdb-min-available" {
+
+  tmp="tests/checks/pod-disruption-budget.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include pdb-min-available --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}  
+}
+
```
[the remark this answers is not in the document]
fahlmant: Is there any documentation for writing these tests?

docs/generated/checks.md
```diff
@@ -410,6 +410,24 @@ strategyTypeRegex: ^(RollingUpdate|Rolling)$
...
+**Remediation**: Change the "Pod Disruption Budget" to have a "MaxUnavailableValue" set greater than 0. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information.
+
+**Template**: [pdb-max-unavailable](templates.md#pod-disruption-budget-configuration-checks)
+## pdb-min-available
+
+**Enabled by default**: Yes
+
+**Description**: Indicates when a DeploymentLike object has a "Pod Disruption Budget" specified with a MinAvailable value that is too restrictive.
+
+**Remediation**: Change the "Pod Disruption Budget" to have a "MinAvailableValue" set to number lower than the Replica count in the related DeploymentLike. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information.
```
dhaus67, line 428: nit:
```suggestion
**Remediation**: Change the "Pod Disruption Budget" to have a "MinAvailableValue" set to a number lower than the Replica count in the related DeploymentLike object. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information.
```

e2etests/bats-tests.sh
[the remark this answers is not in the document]
janisz: only other examples and bats documentation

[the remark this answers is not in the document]
fahlmant: I've built the tests with assertions now 👍

tests/checks/pdb-min-available.yaml
```diff
@@ -0,0 +1,22 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: foo
+  namespace: bar
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      name: cloud-ingress-operator
```
dhaus67, line 22: Missing newline.

dhaus67 (approved):
Just minor things, otherwise LGTM!

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,173 @@
...
+	if pdb.Spec.MinAvailable == nil {
+		return results
+	}
+
+	// Extract the MinAvailable value from the spec and check if it's a number or percentage
+	value, isPercent, err := getIntOrPercentValueSafelyFromString(pdb.Spec.MinAvailable.String())
+	if err != nil {
+		return []diagnostic.Diagnostic{
+			{
+				Message: fmt.Sprintf("PDB has invalid MinAvailable value: %v", err),
```
janisz, line 64: let's wrap this error or use `errors.Wrap` like we do in other places
```suggestion
				Message: fmt.Sprintf("PDB has invalid MinAvailable value: %w", err),
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,173 @@
...
+			},
+		}
+	}
+
+	// Evaluate Deploymet Likes in the lintContext to see if they have MinAvailable set too low
+	deploymentLikes, err := getDeploymentLikeObjects(lintCtx, labelSelector, pdb.Namespace)
+	if err != nil {
+		return []diagnostic.Diagnostic{
+			{
+				Message: fmt.Sprintf("Failed to retrieve deployments matching the PDB's label selector within namespace %s: %v", pdb.Namespace, err.Error()),
```
janisz, line 98: as above
```suggestion
				Message: fmt.Sprintf("Failed to retrieve deployments matching the PDB's label selector within namespace %s: %w", pdb.Namespace, err.Error()),
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,173 @@
...
+	for _, dl := range deploymentLikes {
+		pdbMinAvailable := value
+		replicas, _ := extract.Replicas(dl)
+		if isPercent {
+			// Calulate the actual value of the MinAvailable with respect to the Replica count if a percentage is set
+			pdbMinAvailable = int(math.Floor(float64(replicas) * (float64(value) / float64(100))))
+		}
+		if replicas <= int32(pdbMinAvailable) {
+			results = append(results, diagnostic.Diagnostic{
+				Message: fmt.Sprintf("Deployment %s has replicas less than or equal to the minimum available replicas set by its PDB.", dl.GetName()),
```
janisz, line 112: Something is wrong with this message but I'm not sure what how about something like this
> The current number of replicas for deployment %s is equal to or lower than the minimum number of replicas specified by its PDB.

pkg/templates/pdbminavailable/template.go
janisz: @tremes is right but if we change order (multiplication first) it will work without conversion and floor

```
pdbMinAvailable = (int(replicas) * value) / 100
```

janisz (changes requested):
Just small comments. I think we should use `errors.Wrap` to wrap errors.

tests/checks/pdb-min-available.yaml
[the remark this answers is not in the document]
fahlmant: Can you clarify this a bit? I'm not sure what is being asked here

pkg/templates/pdbmaxunavailable/template.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+const (
+	templateKey           = "pdb-max-unavailable"
+	maxUnavailableZeroMsg = "MaxUnavailable is set to 0"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Pod Disruption Budget Configuration Checks",
```
jholecek-rh, line 21: Both checks have the same strings, resulting in two sections with identical title in the generated `template.md`. How about tailoring the strings to the individual checks?

(I paused here for quite a while trying to figure out more specific strings. Having two separate checks for the two configuration settings makes it more difficult that I thought.)

How about this?

```suggestion
		HumanName:   "No pod disruption allowed - maxUnavailable",
```

pkg/templates/pdbmaxunavailable/template.go
```diff
@@ -0,0 +1,53 @@
...
+		Key:         templateKey,
+		Description: "Flag non-optimal PDB configurations",
```
jholecek-rh, line 23: "Non-optimal" is too vague for what the check does, imho. How about this?

```suggestion
		Description: "Flag PodDisruptionBudgets whose maxUnavailable value prevents pod disruptions.",
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,173 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+const (
+	templateKey = "pdb-min-available"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Pod Disruption Budget Configuration Checks",
```
jholecek-rh, line 29: Similar as above.

```suggestion
		HumanName:   "No pod disruptions allowed - minAvailable",
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,173 @@
...
+		Key:         templateKey,
+		Description: "Flag non-optimal PDB configurations",
```
jholecek-rh, line 31: Similar as above.

```suggestion
		Description: "Flag PodDisruptionBudgets whose minAvailable value prevents pod disruptions.",
```

pkg/templates/pdbminavailable/template.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	for _, dl := range deploymentLikes {
+		pdbMinAvailable := value
+		replicas, _ := extract.Replicas(dl)
+		if isPercent {
+			// Calulate the actual value of the MinAvailable with respect to the Replica count if a percentage is set
+			pdbMinAvailable = int(math.Floor(float64(replicas) * (float64(value) / float64(100))))
+		}
```
jholecek-rh, line 109: Is this necessary? 

IIUC, the only case when `pdbMinAvailable` would be lower than or equal to `replicas` is for `value == 100` which is handled separately above. I think that we have to loop through the deployment-like objects only when `minAvailable` is set to an absolute value.

fahlmant: Sure, I guess we can return here.

pkg/builtinchecks/yamls/pdbs-max-unavailable.yaml
```diff
@@ -0,0 +1,7 @@
+name: "pdb-max-unavailable"
+description: "Indicates when a DeploymentLike object has a \"Pod Disruption Budget\" specified with a MaxUnavailable value that is too restrictive."
+remediation: "Change the \"Pod Disruption Budget\" to have a \"MaxUnavailableValue\" set greater than 0. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information."
```
jholecek-rh, line 3: Would you agree that "prevents disruptions" is more informative than "too restrictive"?

Also adjusting mark-up to be consistent with other checks (they don't seem to denote literal object and field names typographically) and fixing the "MaxUnavailableValue" typo.

```suggestion
description: "Indicates when a deployment-like object has a PodDisruptionBudget specified with a maxUnavailable value that prevents disruptions."
remediation: "Change the PodDisruptionBudget to have maxUnavailable set to a value greater than 0. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information."
```

The description still sounds as if the check reported deployment-like objects rather than PDBs. How about this?

```suggestion
description: "Indicates when a PodDisruptionBudget has a maxUnavailable value that prevents disruptions of deployment-like objects it is applied to."
remediation: "Change the PodDisruptionBudget to have maxUnavailable set to a value greater than 0. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information."
```

pkg/builtinchecks/yamls/pdbs-min-available.yaml
```diff
@@ -0,0 +1,7 @@
+name: "pdb-min-available"
+description: "Indicates when a DeploymentLike object has a \"Pod Disruption Budget\" specified with a MinAvailable value that is too restrictive."
+remediation: "Change the \"Pod Disruption Budget\" to have a \"MinAvailableValue\" set to number lower than the Replica count in the related DeploymentLike. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information."
```
jholecek-rh, line 3: Similar as for the other check.

Option 1:

```suggestion
description: "Indicates when a deployment-like object has a PodDisruptionBudget specified with a minAvailable value that prevents disruptions."
remediation: "Change the PodDisruptionBudget to have minAvailable set to a number lower than the number of replicas in the related deployment-like objects. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information."
```

Option 2 (turning the description around):

```suggestion
description: "Indicates when a PodDisruptionBudget sets a minAvailable value that will always prevent disruptions of pods created by related deployment-like objects."
remediation: "Change the PodDisruptionBudget to have minAvailable set to a number lower than the number of replicas in the related deployment-like objects. Refer to https://kubernetes.io/docs/tasks/run-application/configure-pdb/ for more information."
```

pkg/templates/pdbminavailable/template.go
jholecek-rh: It is necessary indeed. Kubernetes rounds maxUnavailable and minAvailable **up**, **not  down**.

See
* [PDB docs](https://kubernetes.io/docs/tasks/run-application/configure-pdb/#rounding-logic-when-specifying-percentages)
* [Related PDB code](https://github.com/kubernetes/kubernetes/blob/23be9587a0f8677eb8091464098881df939c44a9/pkg/controller/disruption/disruption.go#L539)
* [GetValueFromIntOrPercent implementation](https://github.com/kubernetes/apimachinery/blob/9cb1f71d9bce60a0ec671b70613892c055f8ff49/pkg/util/intstr/intstr.go)
```suggestion
		if isPercent {
			// Calulate the actual value of the MinAvailable with respect to the Replica count if a percentage is set
			pdbMinAvailable = int(math.Ceil(float64(replicas) * (float64(value) / float64(100))))
		}
```

pkg/templates/pdbminavailable/template.go
fahlmant: Due to that:
`fmt.Sprintf does not support error-wrapping directive %w`

pkg/templates/pdbminavailable/template.go
fahlmant: `fmt.Sprintf does not support error-wrapping directive %w`

pkg/templates/pdbminavailable/template.go
janisz: I'm sorry, my bad

janisz (approved):

maorfr:
https://issues.redhat.com/browse/DVO-13

PR #535, opened 2023-04-11
Fix go install for newer golangs

SuperSandro2000:
I was greeted with this error:
```
 ▶ GO111MODULE=on go install golang.stackrox.io/kube-linter/cmd/kube-linter
go: 'go install' requires a version when current directory is not in a module
        Try 'go install golang.stackrox.io/kube-linter/cmd/kube-linter@latest' to install the latest version
```
also setting ``GO11MODULE`` shouldn't be required since a while now.


janisz (approved):

janisz:
@SuperSandro2000 Thank you!

PR #546, opened 2023-04-13
fix/linting for percentage based values

bmiguel-teixeira:
This PR fixes the PDB validation to include support for % based values.

OldBehavior:
maxUnavailable was set to 10% was failing since the field is IntOrString. The IntVal defaults to 0 when the field is of String Values, thus causing false positives.

NewBehaviour:
int and string values are taken into account, parsed into a single integer value for validation.

### DIFF
--- a/pkg/templates/pdbmaxunavailable/template.go
+++ b/pkg/templates/pdbmaxunavailable/template.go
@@ -1,6 +1,8 @@
 package pdbmaxunavailable
 
 import (
+	"fmt"
+
 	"golang.stackrox.io/kube-linter/pkg/check"
 	"golang.stackrox.io/kube-linter/pkg/config"
 	"golang.stackrox.io/kube-linter/pkg/diagnostic"
@@ -9,11 +11,11 @@ import (
 	"golang.stackrox.io/kube-linter/pkg/templates"
 	"golang.stackrox.io/kube-linter/pkg/templates/pdbmaxunavailable/internal/params"
 	pdbV1 "k8s.io/api/policy/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
 const (
-	templateKey           = "pdb-max-unavailable"
-	maxUnavailableZeroMsg = "MaxUnavailable is set to 0"
+	templateKey = "pdb-max-unavailable"
 )
 
 func init() {
@@ -29,24 +31,30 @@ func init() {
 		ParseAndValidateParams: params.ParseAndValidate,
 		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
 			return func(lintCtx lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
-
 				// Get the PDB provided
 				pdb, ok := object.K8sObject.(*pdbV1.PodDisruptionBudget)
 				if !ok {
 					return nil
 				}
 
-				// If MaxUnavailable is set to 0 (both int and string)
-				// return as no other checks will uniquely apply
-				if pdb.Spec.MaxUnavailable != nil && pdb.Spec.MaxUnavailable.IntVal == 0 {
+				if pdb.Spec.MaxUnavailable == nil {
+					return []diagnostic.Diagnostic{}
+				}
+
+				maxUnavailable, err := intstr.GetScaledValueFromIntOrPercent(pdb.Spec.MaxUnavailable, 100, false)
+				if err != nil {
 					return []diagnostic.Diagnostic{{
-						Message: maxUnavailableZeroMsg,
-					},
-					}
+						Message: fmt.Sprintf("maxUnavailable has invalid value [%s]", pdb.Spec.MaxUnavailable),
+					}}
+				}
+
+				if maxUnavailable == 0 {
+					return []diagnostic.Diagnostic{{
+						Message: "MaxUnavailable is set to 0",
+					}}
 				}
 
 				return []diagnostic.Diagnostic{}
-
 			}, nil
 		}),
 	})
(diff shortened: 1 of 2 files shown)


janisz:
@fahlmant  PTAL

jholecek-rh:
Since the MR is importing `k8s.io/apimachinery/pkg/util/intstr`, wouldn't it be better to refactor the code to use higher-level functions offered by that package, e.g. [getIntOrPercentValueSafely](https://github.com/kubernetes/apimachinery/blob/8d1258da8f386b809d312cdda316366d5612f54e/pkg/util/intstr/intstr.go#L207)?

pkg/templates/pdbmaxunavailable/template.go
```diff
@@ -29,24 +33,40 @@ func init() {
...
 				}
 
-				// If MaxUnavailable is set to 0 (both int and string)
-				// return as no other checks will uniquely apply
-				if pdb.Spec.MaxUnavailable != nil && pdb.Spec.MaxUnavailable.IntVal == 0 {
-					return []diagnostic.Diagnostic{{
-						Message: maxUnavailableZeroMsg,
-					},
+				if pdb.Spec.MaxUnavailable == nil {
+					if pdb.Spec.MaxUnavailable.IntVal == 0 {
```
tremes, line 43: If the first condition is satisfied then the second one will fail with panic.

pkg/templates/pdbmaxunavailable/template.go
```diff
@@ -29,24 +33,40 @@ func init() {
...
+						return []diagnostic.Diagnostic{{
+							Message: "MaxUnavailable is undefined",
```
tremes, line 45: I think this is "undesired" diagnostic. I mean we would like to report when `MaxUnavailable` is 0, otherwise do nothing.

mklika:
Hi @bmiguel-teixeira! 

My name is Martin and I work at Red Hat along with Tomas. We're quite eager to see these changes merged as we bellieve PDB check could be quite helpful for our SRE organization. 

Would you mind letting me know your estimate in terms of when would you have time look at comments from Tomas and fixing the code? 

I'd like to also offer our help finishing this up should it be your preference due to other priorities you might have.

Many thanks in advance for letting me know

Martin Klika

pkg/templates/pdbmaxunavailable/template.go
bmiguel-teixeira: done.

pkg/templates/pdbmaxunavailable/template.go
bmiguel-teixeira: still have a bit of mixed feeling on this one but a "missing" maxUnavailable is valid. skipping validation on nil. please take a look

pkg/templates/pdbmaxunavailable/template.go
tremes: Thank you

pkg/templates/pdbmaxunavailable/template.go
tremes: yes it's valid. Thanks. Looks good.

tremes:
Looks good to me. Thank you @bmiguel-teixeira .

pkg/templates/pdbmaxunavailable/template_test.go
```diff
@@ -28,11 +28,49 @@ func (p *PDBTestSuite) SetupTest() {
 	p.ctx = mocks.NewMockContext()
 }
 
-func (p *PDBTestSuite) TestPDBMaxUnavailableZero() {
+func (p *PDBTestSuite) TestPDBMaxUnavailableZeroAsInteger() {
```
janisz, line 31: How about converting this tests to table driven tests?
```go
func (p *PDBTestSuite) TestPDB() {
	testCases := []struct {
		Name           string
		Message        string
		MaxUnavailable intstr.IntOrString
	}{
		{
			Name:           "Invalid Value String",
			Message:        "maxUnavailable has invalid value [invalid-value]",
			MaxUnavailable: intstr.FromString("invalid-value"),
		},
		{
			Name:           "Invalid Value String",
			Message:        "maxUnavailable has invalid value [0]",
			MaxUnavailable: intstr.FromString("0"),
		},
		{
			Name:           "One As String Percentage",
			MaxUnavailable: intstr.FromString("1%"),
		},
		{
			Name:           "One As Integer",
			MaxUnavailable: intstr.FromInt(1),
		},
		{
			Name:           "Zero As String Percentage",
			Message:        "MaxUnavailable is set to 0",
			MaxUnavailable: intstr.FromString("0%"),
		},
		{
			Name:           "Zero As Integer",
			Message:        "MaxUnavailable is set to 0",
			MaxUnavailable: intstr.FromInt(0),
		},
	}

	for _, tc := range testCases {
		tc := tc
		p.Run(tc.Name, func() {
			p.ctx.AddMockPodDisruptionBudget(p.T(), "test-pdb")
			p.ctx.ModifyPodDisruptionBudget(p.T(), "test-pdb", func(pdb *pdbV1.PodDisruptionBudget) {
				pdb.Spec.MaxUnavailable = &tc.MaxUnavailable
			})

			expected := map[string][]diagnostic.Diagnostic{}
			if tc.Message != "" {
				expected = map[string][]diagnostic.Diagnostic{
					"test-pdb": {{Message: tc.Message}},
				}
			}

			p.Validate(p.ctx, []templates.TestCase{
				{
					Param:                    params.Params{},
					Diagnostics:              expected,
					ExpectInstantiationError: false,
				},
			})
		})
	}
}
```

bmiguel-teixeira: make sense. done

janisz (approved):

PR #553, opened 2023-04-24
Update Docker pre-commit-hook to 0.6.3 image

m3adow:


janisz (approved):

janisz:
Thanks! I think we should run precommit as a part of our CI

PR #595, opened 2023-07-03
updated pre commit hook docker image version

PedroSilva9:
pre-commit with docker image was returning the following error: Error response from daemon: manifest for stackrox/kube-linter:0.6.3 not found: manifest unknown: manifest unknown

The correct syntax would be `stackrox/kube-linter:v0.6.3`. Since a new version was released I took the liberty to update pre-commit one's.


janisz (approved):

janisz:
I think we should add a post release step to create PR that will automatically update version once it's released.

PR #614, opened 2023-08-21
[Docs] make text match example for custom required-annotation

afirth:
Previously link was label but description and example were for annotation. Generalize and correct


janisz (approved):

janisz:
@afirth Thank you!

PR #661, opened 2023-11-17
Create new check for liveness port

charlesoconor:
Want to ensure the port is correctly set for liveness probe since it's easy to make a typo when people are doing copy pasta. This will let you know it's failing airier and not have to wait for a failed deployment.

Want to do this also for startup and readiness checks. Assume it make sense to have them as other checks and was planning on doing that in a follow up.


pkg/builtinchecks/yamls/liveness-port.yaml
```diff
@@ -0,0 +1,9 @@
+name: "liveness-port"
+description: "Indicates when containers have a liveness probe to a port the container didn't open."
```
janisz, line 2:
```suggestion
description: "Indicates when containers have a liveness probe to a not exposed port."
```

https://github.com/garethr/kubernetes-json-schema/blob/8aa572595b98d73b2b9415ca576f78e163381b10/v1.14.0-standalone/deployment.json#L2201

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+
+const templateKey = "liveness-http-port"
+
+var sentinal = struct{}{}
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "Liveness Port Not Open",
+		Key:         templateKey,
+		Description: "Flag containers have a http liveness prop with a port they didn't open.",
```
janisz, line 25: Something like this will be better
```suggestion
		Description: "Flag containers with an HTTP liveness probe to not exposed port.",
```

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+				}
+
+				httpProbe := container.LivenessProbe.ProbeHandler.HTTPGet
+				if httpProbe == nil {
+					return nil
+				}
+
+				ports := map[intstr.IntOrString]struct{}{}
+				for _, port := range container.Ports {
+					ports[intstr.FromInt(int(port.ContainerPort))] = sentinal
```
janisz, line 44:
```suggestion
					ports[intstr.FromInt32(int(port.ContainerPort))] = sentinal
```

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+	"golang.stackrox.io/kube-linter/pkg/templates"
+	"golang.stackrox.io/kube-linter/pkg/templates/livenessport/internal/params"
+	"golang.stackrox.io/kube-linter/pkg/templates/util"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+const templateKey = "liveness-http-port"
+
+var sentinal = struct{}{}
```
janisz, line 19:
```suggestion
var sentinel = struct{}{}
```

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+					return nil
+				}
+
+				httpProbe := container.LivenessProbe.ProbeHandler.HTTPGet
+				if httpProbe == nil {
+					return nil
+				}
+
+				ports := map[intstr.IntOrString]struct{}{}
+				for _, port := range container.Ports {
```
janisz, line 43: Do we need to skip `port.Protocol != http`?

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
+			return util.PerNonInitContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
+				if container.LivenessProbe == nil {
+					return nil
+				}
+
+				httpProbe := container.LivenessProbe.ProbeHandler.HTTPGet
```
janisz, line 37: How about extending this check for TCP ?

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+
+				ports := map[intstr.IntOrString]struct{}{}
+				for _, port := range container.Ports {
+					ports[intstr.FromInt(int(port.ContainerPort))] = sentinal
+					ports[intstr.FromString(port.Name)] = sentinal
+				}
+
+				if _, ok := ports[httpProbe.Port]; !ok {
+					return []diagnostic.Diagnostic{{
+						Message: fmt.Sprintf("container %q does not have an open port %s", container.Name, httpProbe.Port.String()),
```
janisz, line 50:
```suggestion
						Message: fmt.Sprintf("container %q does not expose port %s", container.Name, httpProbe.Port.String()),
```

pkg/templates/livenessport/template.go
```diff
@@ -0,0 +1,57 @@
...
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(_ params.Params) (check.Func, error) {
+			return util.PerNonInitContainerCheck(func(container *v1.Container) []diagnostic.Diagnostic {
+				if container.LivenessProbe == nil {
+					return nil
+				}
+
+				httpProbe := container.LivenessProbe.ProbeHandler.HTTPGet
+				if httpProbe == nil {
+					return nil
```
janisz, line 39: how about testing this case?

tests/checks/invalid-target-ports.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: v1
+napiVersion: v1
```
janisz, line 1:
```suggestion
apiVersion: v1
```

janisz (changes requested):
Nice. I've left some comments. I think we should use the same nomencalture as k8s so ports are `exposed` not `opened`. I think we should also support `TCP` probes not only `http`.

pkg/templates/livenessport/template.go
charlesoconor: Good idea adding a check since UDP sockets would fail. 

```
if port.Protocol != "" && port.Protocol != "TCP" {
    continue
}
```

pkg/templates/livenessport/template.go
charlesoconor: Good idea adding a check since UDP sockets would fail. 

```
if port.Protocol != "" && port.Protocol != "TCP" {
    continue
}
```

charlesoconor:
Thanks for the review! Have updated with the requested changes.

charlesoconor:
@janisz, is there anything else you want me to change? Take another look when you have a chance.

janisz (approved):

janisz:
@charlesoconor Thank you. I'm sorry I missed your comment, you can re-request review to speed up the process as I'll see this PR on my https://github.com/pulls board

Issue #576, opened 2023-05-31
[BUG] Template links are broken


janisz:
When navigating to https://docs.kubelinter.io/#/generated/checks?id=no-node-affinity and clicking template we are getting redirected to wrong page resulting in 404

```diff
+ https://docs.kubelinter.io/#/templates?id=node-affinity
- https://docs.kubelinter.io/#/generated/templates?id=anti-affinity-not-specified
```

PR #439, opened 2022-10-20
Fix issue #431

juandspy:
Fix #431

PR #662, opened 2023-11-21
Docs: Fixed broken template links, fixed script versions

gaurav-nelson:
Fixes https://github.com/stackrox/kube-linter/issues/576

I see that `generated` was removed in https://github.com/stackrox/kube-linter/pull/439 which I am think was caused by changes in `docsify` library. Therefore, I've updated the script tags to load a fixed version so that it doesn't happen again.


docs/generated/checks.md
```diff
@@ -10,7 +10,7 @@ KubeLinter includes the following built-in checks:
 
 **Remediation**: Where possible, remove create access to pod objects in the cluster.
 
-**Template**: [access-to-resources](templates.md#access-to-resources)
+**Template**: [access-to-resources](generated/templates.md#access-to-resources)
```
janisz, line 13: This link is broken. If you go to https://github.com/stackrox/kube-linter/blob/553365517c108e7cb200af7d98527b8716e236ac/docs/generated/checks.md and click that link it will give you https://github.com/stackrox/kube-linter/blob/553365517c108e7cb200af7d98527b8716e236ac/docs/generated/generated/templates.md#access-to-resources that does not exists

gaurav-nelson: I'm not sure how GitHub is handling the links. I've built the site locally with docsify and used `docsify serve ./docs` command to check. 

For example, go to https://stackblitz.com/~/github.com/gaurav-nelson/kube-linter/tree/fix-template-links and run `npm i docsify-cli` and then `docsify server ./docs` commands to check.

janisz: Ah, that's how it's done. How about keeping it working on github and docksify by adding `relativePath: true,` in 
https://github.com/stackrox/kube-linter/blob/553365517c108e7cb200af7d98527b8716e236ac/docs/index.html#L16-L19
Then we only need to change `index.html`

gaurav-nelson: I tried this locally. It fixes the check and template links. But if you click another topic, for example, Using KubeLinter it tries to open `http://localhost:3000/#/generated/using-kubelinter` 404.

janisz: Good catch! So we need to change index and https://github.com/stackrox/kube-linter/blob/553365517c108e7cb200af7d98527b8716e236ac/docs/_sidebar.md to use absolute paths. Then it will not work on github and IDE but probably it's fine as nobody will use this file

janisz: Hmm, it's working for me with this patch
```diff
diff --git a/docs/_sidebar.md b/docs/_sidebar.md
index 3457af1..3412f76 100644
--- a/docs/_sidebar.md
+++ b/docs/_sidebar.md
@@ -1,5 +1,5 @@
 * [Introduction](/)
-* [Using KubeLinter](using-kubelinter.md)
-* [Configuring KubeLinter](configuring-kubelinter.md)
-* [KubeLinter checks](generated/checks.md)
-* [KubeLinter templates](generated/templates.md)
\ No newline at end of file
+* [Using KubeLinter](/using-kubelinter.md)
+* [Configuring KubeLinter](/configuring-kubelinter.md)
+* [KubeLinter checks](/generated/checks.md)
+* [KubeLinter templates](/generated/templates.md)
diff --git a/docs/index.html b/docs/index.html
index dc7b2eb..deab721 100644
--- a/docs/index.html
+++ b/docs/index.html
@@ -27,6 +27,7 @@
         errorText  : 'Error',
         successText: 'Copied'
       },
+      relativePath: true,
       'flexible-alerts': {
         style: 'flat'
       },

```

janisz:
@gaurav-nelson I've fixed this on your PR. Could you check. It's working for me

janisz (approved):

Issue #575, opened 2023-05-29
[BUG] Ignore-paths parameter can not take effect in my project


darmenliu:
**System info:**
 - OS: [e.g. Linux? MaxOS? Windows?]
 - Linux

**Describe the bug**
I have a project and I want use kube-linter to san the helm chart, but I failed to ignore test path to use ignore-paths parameter.

```
myproject$ ls
templates  kube_linter.yaml  value.yaml

myproject/templates$ ls
test  others.yaml

myproject$  kube-linter lint ./ --config ./kube_linter.yaml --ignore-paths ./templates/test/
templates/test/robot.yaml: (object: <no namespace>/robot-pod /v1, Kind=Pod) found matching serviceAccount ("") (check: default-service-account, remediation: Create a dedicated service account for your pod. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/ for details.)

templates/test/robot.yaml: (object: <no namespace>/robot-pod /v1, Kind=Pod) container "robot-pod-container" does not specify a liveness probe (check: no-liveness-probe, remediation: Specify a liveness probe in your container. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ for details.)

templates/test/robot.yaml: (object: <no namespace>/robot-pod /v1, Kind=Pod) container "robot-pod-container" is not set to runAsNonRoot (check: run-as-non-root, remediation: Set runAsUser to a non-zero number and runAsNonRoot to true in your pod or container securityContext. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for details.)

templates/test/robot.yaml: (object: <no namespace>/robot-pod /v1, Kind=Pod) resource specifies unsafe sysctl "net.ipv4.ip_local_port_range". (check: unsafe-sysctls, remediation: Ensure container does not allow unsafe allocation of system resources by removing unsafe sysctls configurations. For more details see https://kubernetes.io/docs/tasks/administer-cluster/sysctl-cluster/ https://docs.docker.com/engine/reference/commandline/run/#configure-namespaced-kernel-parameters-sysctls-at-runtime.)

```
Use this command to scan my project seems my test file still was reported by kube-linter, hope some one can help me what's wrong for my command, and I also try configure this with configuration like:
checks:
ignorePaths:
- ./templates/test/robot.yaml

but I still failed.

**To Reproduce**
1. Add a test directory in a project of helm charts.
[...]

PR #667, opened 2023-11-23
feat: ignore templates in Chart dependencies based on ignore paths

Trojan295:
Solves #575


janisz (commented):
Nice, could you add tests to validate it's working as expected in the issue?

Trojan295:
@janisz I added a test.

I also saw that the `ignorePaths` are always passed to the `CreateContexts` in absolute path (), while the args might be absolute or relative. You can see that by comparing, which yield a different result on `main`:
```bash
kube-linter lint $PWD/tests/testdata --ignore-paths $PWD/tests/testdata

kube-linter lint tests/testdata --ignore-paths $PWD/tests/testdata
```

The PR also addresses this.

janisz:
Thank you!
I'm sorry for a long delay.

Issue #637, opened 2023-10-10
[BUG] Linter reporting a low number of replicas for deployments with PDB and HPA minReplicas


gerodriguez-fan:
**System info:**
 - OS: Linux

**Describe the bug**
When linting a deployment that contains a PodDisruptionBudget with `maxUnavailable: 1` and an HPA with `minReplicas: 4` the tool reports: 

> The current number of replicas for deployment ecomprod-master-c1-sink-vector-worker is equal to or lower than the minimum number of replicas specified by its PDB.

It request to update the `minAvailable` field on the PDB which cannot be set if the `maxUnavailable` field is used.
This is a common scenario, we want an HPA to start with a certain number of replicas and we do not want it to be disrupted more than 1 pod at the time.

**To Reproduce**
run kube-linter on any deployment with a PodDisruptionBudget with maxUnavailable of 1 and and HPA with minReplicas:4

**Sample YAML input**

**Expected behavior**
The linter should not trigger a failure as this is a common scenario.

janisz:
@gerodriguez-fan Thank you for reporting this. Could you please add an example yaml (you can use our test yaml`s as a base) just to make sure we are talking about same thing.

From the description the problem could be the replicas we are using. So we are comparing replicas not `minReplicas` in PDB.

https://github.com/stackrox/kube-linter/blob/e2c0cdbf13ea811a864d857c50d961f370e5732c/pkg/templates/pdbminavailable/template.go#L105-L112

@gerodriguez-fan Would you like to work on this issue?

jpedrobf:
I'm having a similar issue even when I use a percentage value on `minAvailable`. My deployment spec doesn't have `.spec.replicas` declared as I use the HPA min/max parameters and any value I put on the pdb ends up with an error.

my pdb spec: 
```yaml
spec:
  minAvailable: 75%
  selector:
    matchLabels:
      app: example
```

my deployment spec:
```yaml
spec:
  revisionHistoryLimit: 3
  progressDeadlineSeconds: 600
  strategy:
    rollingUpdate:
      maxSurge: 10%
      maxUnavailable: 0
    type: RollingUpdate
  selector:
    matchLabels:
      app: example
```

---

What I've noticed: 
- This started to happen when I updated the pdb apiVersion from policy/v1beta1 to policy/v1.
[...]

PR #688, opened 2024-01-08
Fix: pdb-min-available when Replica number is controlled via HPA

jpedrobf:
I believe this solves #637. 

The commit history might be a little messy but it was made with heart 💚

### DIFF
--- a/pkg/templates/pdbminavailable/template.go
+++ b/pkg/templates/pdbminavailable/template.go
@@ -15,6 +15,7 @@ import (
 	"golang.stackrox.io/kube-linter/pkg/objectkinds"
 	"golang.stackrox.io/kube-linter/pkg/templates"
 	"golang.stackrox.io/kube-linter/pkg/templates/pdbminavailable/internal/params"
+	autoscalingV2 "k8s.io/api/autoscaling/v2"
 	pdbV1 "k8s.io/api/policy/v1"
 	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -100,14 +101,32 @@ func minAvailableCheck(lintCtx lintcontext.LintContext, object lintcontext.Objec
 		}
 	}
 
+	// If there are no Replicas set on the Deployment Like, check if there is an HPA with a minReplicas set, and use that value, Else fail
+	if _, ok := extract.Replicas(object.K8sObject); ok {
+		// Get the HPA provided
+		hpa, ok := object.K8sObject.(*autoscalingV2.HorizontalPodAutoscaler)
+		if !ok {
+			return nil
+		}
+		// Ensure that the hpa scaleTargetRef is the deployment name
+		if hpa.Spec.ScaleTargetRef.Name != object.GetK8sObjectName().Name {
+			return nil
+		}
+		// If the HPA has a minReplicas set, use that value, else fail
+		if hpa.Spec.MinReplicas != nil {
+			value = int(*hpa.Spec.MinReplicas)
+		} else {
+			return []diagnostic.Diagnostic{
+				{
+					Message: fmt.Sprintf("Deployment %s has no replicas set, and the HPA %s has no minReplicas set", object.GetK8sObjectName().Name, hpa.GetName()),
+				},
+			}
+		}
+	}
+
 	for _, dl := range deploymentLikes {
 		pdbMinAvailable := value
 		replicas, _ := extract.Replicas(dl)
-		replicasHpa, found := extract.HPAMinReplicas(object.K8sObject)
-		// if replicas is 0, then use the value from the HPA if it exists
-		if replicas == 0 && found {
-			replicas = replicasHpa
-		}
 		if isPercent {
 			// Calulate the actual value of the MinAvailable with respect to the Replica count if a percentage is set
 			pdbMinAvailable = int(math.Ceil(float64(replicas) * (float64(value) / float64(100))))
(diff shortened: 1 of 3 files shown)


pkg/templates/pdbminavailable/template.go
```diff
@@ -103,6 +104,33 @@ func minAvailableCheck(lintCtx lintcontext.LintContext, object lintcontext.Objec
...
+					replicas = *hpa.Spec.MinReplicas
+				} else {
+					return []diagnostic.Diagnostic{
+						{
+							Message: fmt.Sprintf("Deployment %s has no replicas set, and the HPA has no minReplicas set or doesn't exist", dl.GetName()),
+						},
+					}
+				}
+			}
+		}
```
janisz, line 133: I think this part should be extracted to a function

pkg/templates/pdbminavailable/template.go
```diff
@@ -103,6 +104,33 @@ func minAvailableCheck(lintCtx lintcontext.LintContext, object lintcontext.Objec
...
+		if int(replicas) == 1 {
+			hpaList, err := getHorizontalPodAutoscalers(lintCtx, pdb.Namespace)
+			if err != nil {
+				break
+			}
+			for _, hpa := range hpaList {
+				// Select from the list of HPAs one that have scaleTargetRef that matches the deployment name
+				hpa, ok := hpa.(*autoscalingV2.HorizontalPodAutoscaler)
+				if !ok {
+					return nil
```
janisz, line 116: should we `continue` here instead returning?

pkg/templates/pdbminavailable/template.go
```diff
@@ -103,6 +104,33 @@ func minAvailableCheck(lintCtx lintcontext.LintContext, object lintcontext.Objec
...
+				}
+				// Ensure that the hpa scaleTargetRef is the deployment name
+				if hpa.Spec.ScaleTargetRef.Name != dl.GetName() {
+					continue
+				}
+				// If the HPA has a minReplicas set, use that value as the deployment replica, else fail
+				if hpa.Spec.MinReplicas != nil {
+					replicas = *hpa.Spec.MinReplicas
+				} else {
```
janisz, line 125: If we `continue` in the `if` we can remove `else` and reduce indentation.

jpedrobf: I think I've implemented the corrections stated, but I think this one is an opportunity to deliver a different, possible, error. 

With the else statement, we cover the case where there's no HPA available and the deployment spec.replicas is nil

Though this scenario won't happen right now as the `extract pod_spec Replicas` function currently enforces a return 1, that could change (I've actually tried to change that but its tied to other tests), as it is possible to have nil replicas. 

Not sure what is the best approach here. To YAGNI or to not YAGNI?

jpedrobf:
@janisz, giving an update: @jvtartaglia helped a lot in refining the code to avoid nested `for` inside the code and reduce complexity. If there's any other feedback, we're happy to follow up on it 😄

dineflansouza (approved):

pkg/templates/pdbminavailable/template.go
```diff
@@ -15,6 +15,7 @@ import (
 	"golang.stackrox.io/kube-linter/pkg/objectkinds"
 	"golang.stackrox.io/kube-linter/pkg/templates"
 	"golang.stackrox.io/kube-linter/pkg/templates/pdbminavailable/internal/params"
+	autoscalingV2 "k8s.io/api/autoscaling/v2"
```
janisz, line 18: We need to support all versions :( 
There is a method in extract package that could be useful here although I think we need another one to extract target ref name

https://github.com/stackrox/kube-linter/blob/809459200f8a4f2a4cfbf4f9598090ab73bbeeec/pkg/extract/hpa_spec.go#L11-L25

```go
// HPAScaleTargetRefName extracts Spec.ScaleTargetRef.Name
func HPAScaleTargetRefName(obj k8sutil.Object) (string, bool) {
	switch hpa := obj.(type) {
	case *autoscalingV2Beta1.HorizontalPodAutoscaler:
		return hpa.Spec.ScaleTargetRef.Name, true
	case *autoscalingV2Beta2.HorizontalPodAutoscaler:
		return hpa.Spec.ScaleTargetRef.Name, true
	case *autoscalingV2.HorizontalPodAutoscaler:
		return hpa.Spec.ScaleTargetRef.Name, true
	case *autoscalingV1.HorizontalPodAutoscaler:
		return hpa.Spec.ScaleTargetRef.Name, true
	default:
		return "", false
	}
}
```

janisz (commented):
LGTM except one last comment about supported versions.
Also if you like it will be great to have a e2e test for that with different autoscaling version .e.g: that will pass on this PR

```
---yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: baz
  namespace: bar
spec:
  minAvailable: 1
  selector:
    matchLabels:
      name: example
---
apiVersion: autoscaling/v2beta1
kind: HorizontalPodAutoscaler
metadata:
  name: app
  namespace: bar
spec:
  minReplicas: 4
  maxReplicas: 100
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: foo
    namespace: bar
  targetCPUUtilizationPercentage: 85
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: foo
  namespace: bar
spec:
  selector:
    matchLabels:
      name: example
  template:
    metadata:
      labels:
        name: example
 ```

janisz (approved):

janisz:
@jpedrobf Thank you!

PR #689, opened 2024-01-08
Extend liveness-port logic for two new checks around startup and readiness

charlesoconor:
Also added a check for the new `grpc` check that has been added in newer versions of k8s.


charlesoconor:
@janisz Sorry to bug you. This PR passed with go 1.21 as the toolchain. Initially, I pushed this up without thinking since it had that installed and I had to run `go mod tidy` to get the test to behave. I reverted that change, and it's now failing in CI. Assumed you wouldn't want to push through a go upgrade with a feature PR.  Any advice on how to get this to run correctly in CI? Tests pass locally as well as in CI under 1.21.

janisz (approved):

janisz:
@charlesoconor I've run `go mod tidy` for you

charlesoconor:
Thank you! Anything I need to change before a merge.

janisz:
@charlesoconor Thanks! It will be released soon!

PR #692, opened 2024-01-11
comment suggestion for error handling

blondon1:


janisz:
can you elaborate? I thought both options have the same effect when unwrapping but I agree that we can drop this dependency and use native go wrapping.

Issue #691, opened 2024-01-10
[FEATURE_REQUEST] Common target platforms (ie: linux/arm64)


hilariocoelho:
**Description of the problem/feature request**
It would be nice if there was a container for the most common platforms:
- linux/amd64
- linux/arm64

Additionally, the github release would have the artifacts build for the following platforms:
- linux/amd64
- linux/arm64
- darwin/amd64
- darwin/arm64

**Description of the existing behavior vs. expected behavior**
There is no linux/arm64 docker image

Issue #404, opened 2022-09-02
[FEATURE_REQUEST] Publish a mac arm binary 


charlesoconor:
It would be great to publish a prebuilt version of this tool for mac, darwin, arm. The old intel macs are harder to get you're hands on these days. My current workflow is using bazel to pull in the prebuilt and I have a custom-built artifact for people with arm macs. 

I can try to add this functionality but it will be a little before I can get my hands on some hardware to test it.

janisz:
Hey, this makes sense. We need to:
1. Add `GOOS=darwin GOARCH=arm GOARM=7 CGO_ENABLED=0 ` in makefile binary generation 
https://github.com/stackrox/kube-linter/blob/7553e83b18012e5d0ee60a548bc3af356089433b/Makefile#L84
2. Handle arch in the build script (maybe we can handle it by concatenating `$GOOS` with `$GOARCH` 🤔 ) 
https://github.com/stackrox/kube-linter/blob/b614a5d808aaead38a4b92553a2f3fb738ec28db/scripts/go-build.sh#L53
3. Handle new binary in release script
https://github.com/stackrox/kube-linter/blob/8ef2f4970e5a21e2fc559183bfdb132f532cc1c1/.github/workflows/build.yaml#L288

@charlesoconor I can handle testing. I think I know somebody with Mac M1 who could run couple of commands for me 😄

adam-thomas-privitar:
This is still a problem, not sure why the PR was closed. Frustrating as so many of us have arm macbooks in 2023.

charlesoconor:
I got side tracked since I've gotten all of the devs at my company to run rosetta for other reasons. The PR was closed since it wasn't ready meant to open it onto my own fork to check how the actions would work. Probably not to much more work is needed since go is very good a cross-compilation. I or someone needs to find the time.

chulkilee:
For now I'm using brew to install kube-linter - but I'd like to install it from asdf plugin using the github release..

Seems like the release file name convention is changing as well  (#380) so let's take it as an opportunity to add arch to the file name (or we may use amd64 as the old name)

(I don't know the status of "fat" go binary to support multiple arch)

loeffel-io:
+1

PR #701, opened 2024-01-24
Publish arm64 binaries for linux/mac/windows

charlesoconor:
Fixes: #691
Fixes: #404


.github/workflows/build.yaml
```diff
@@ -337,28 +343,49 @@ jobs:
...
         uses: gfreezy/upload-release-asset@v1.0.2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           release_id: ${{ steps.release_drafter.outputs.id }}
           upload_url: ${{ steps.release_drafter.outputs.upload_url }}
-          asset_path: linux/kube-linter-linux
-          asset_name: kube-linter-linux
+          asset_path: linux/amd64/kube-linter-linux-amd64
+          asset_name: kube-linter-linux-amd64
```
janisz, line 355: this will be a breaking change all users who download with a script need to update

charlesoconor: Do you want to upload a `kube-linter-linux` as well as a `kube-linter-linux-amd64` for backwards compatibility?

scripts/go-build.sh
```diff
@@ -49,8 +49,9 @@ fi
 ldflags=(-s -w "${x_defs[@]}")
 
 [[ -n "${GOOS}" ]] || die "GOOS must be set"
+[[ -n "${GOARCH}" ]] || die "GOARCH must be set"
 bin_name="$(basename "$main_srcdir")"
-output_file="bin/${GOOS}/${bin_name}"
+output_file="bin/${GOOS}/${GOARCH}/${bin_name}"
```
charlesoconor, line 54: Could change this to be part of the file if that makes more sense.

```suggestion
output_file="bin/${GOOS}/${bin_name}-${GOARCH}"
```

Makefile
```diff
@@ -81,11 +83,14 @@ generated-srcs: go-generated-srcs generated-docs
...
 build: source-code-archive
-	@CGO_ENABLED=0 GOOS=darwin scripts/go-build.sh ./cmd/kube-linter
-	@CGO_ENABLED=0 GOOS=linux scripts/go-build.sh ./cmd/kube-linter
-	@CGO_ENABLED=0 GOOS=windows scripts/go-build.sh ./cmd/kube-linter
+	@CGO_ENABLED=0 GOARCH=amd64 GOOS=darwin scripts/go-build.sh ./cmd/kube-linter
+	@CGO_ENABLED=0 GOARCH=arm64 GOOS=darwin scripts/go-build.sh ./cmd/kube-linter
+	@CGO_ENABLED=0 GOARCH=amd64 GOOS=linux scripts/go-build.sh ./cmd/kube-linter
+	@CGO_ENABLED=0 GOARCH=arm64 GOOS=linux scripts/go-build.sh ./cmd/kube-linter
+	@CGO_ENABLED=0 GOARCH=amd64 GOOS=windows scripts/go-build.sh ./cmd/kube-linter
+	@CGO_ENABLED=0 GOARCH=arm64 GOOS=windows scripts/go-build.sh ./cmd/kube-linter
```
janisz, line 91: I think we should split this similar to how it's done  https://github.com/stackrox/stackrox/blob/82304c1669be79a3049d886f7e46145de4004a6c/Makefile#L390-L402

.github/workflows/build.yaml
janisz: I think we should keep it backward compatible for a one or two releases

scripts/go-build.sh
janisz: I think we can change it

janisz:
How about usign goreleaser for this https://github.com/stackrox/kube-linter/pull/720

charlesoconor:
> How about usign goreleaser for this #720

It makes sense. I can close this out.

janisz:
@charlesoconor You may like to see 
- #725

PR #703, opened 2024-01-26
feature: add support to keda scaledobject v1alpha1

jonathanmdr:
I believe it provides support to pdb-min-available when the Replica number is controlled via Keda ScaledObject

PR #711, opened 2024-01-31
feature: add support to keda scaledobject v1alpha1

jonathanmdr:
I believe it provides support to pdb-min-available when the Replica number is controlled via Keda ScaledObject

I closed erroneously this PR: https://github.com/stackrox/kube-linter/pull/703


pkg/extract/hpa_spec.go
```diff
@@ -43,6 +46,8 @@ func HPAScaleTargetRefName(obj k8sutil.Object) (string, bool) {
 		return hpa.Spec.ScaleTargetRef.Name, true
 	case *autoscalingV1.HorizontalPodAutoscaler:
 		return hpa.Spec.ScaleTargetRef.Name, true
+	case *kedaV1Alpha1.ScaledObject:
+		return hpa.Spec.ScaleTargetRef.Name, true
```
janisz, line 50: I this line covered by a unit test?

janisz (changes requested):
Nice work! On the first look everything is fine but on second there is something missing. 
1. You correctly added keda to `extract` but it's never called in unit test
2. Could you add good/bad tests to verify it's really working
3. You need to register keda scheme in https://github.com/stackrox/kube-linter/blob/e0225f476a85c4533eab934105c63f44e9d95a50/pkg/lintcontext/parse_yaml.go#L47

No related to this PR but we need to use `--fail-on-invalid-resource` in e2e as in your case it will highlight the error

```
pdb-min-available.yaml: (object: <no namespace>/ /, Kind=) failed to decode: no kind "ScaledObject" is registered for version "keda.sh/v1alpha1" in scheme "pkg/runtime/scheme.go:100" (check: failed-to-load-object, remediation: Confirm that the file is accessible and is valid k8s yaml.)
```

pkg/extract/hpa_spec.go
jonathanmdr: I adjusted the test code to cover this!

jonathanmdr:
> Nice work! On the first look everything is fine but on second there is something missing.
> 
> 1. You correctly added keda to `extract` but it's never called in unit test
> 2. Could you add good/bad tests to verify it's really working
> 3. You need to register keda scheme in https://github.com/stackrox/kube-linter/blob/e0225f476a85c4533eab934105c63f44e9d95a50/pkg/lintcontext/parse_yaml.go#L47
> 
> No related to this PR but we need to use `--fail-on-invalid-resource` in e2e as in your case it will highlight the error
> 
> ```
> pdb-min-available.yaml: (object: <no namespace>/ /, Kind=) failed to decode: no kind "ScaledObject" is registered for version "keda.sh/v1alpha1" in scheme "pkg/runtime/scheme.go:100" (check: failed-to-load-object, remediation: Confirm that the file is accessible and is valid k8s yaml.)
> ```

Well noted!

I adjusted my implementation and unit tests writing good/bad scenarios for this functionality!

If there's any other feedback, we're happy to follow up on it 😄

janisz (approved):

janisz:
Nice, thank you. Could you rebase?

jonathanmdr:
> Nice, thank you. Could you rebase?

Sure, rebase applied!

janisz:
@jonathanmdr  Thanks! In the meantime dependabot updated k8s so I updated your PR.

Issue #695, opened 2024-01-23
Change unset-memory-requirements check


janisz:
Currently the `unset-memory-requirements` requires both requests and limits. This is redundant as request if not set will be copied from limits. The memory differs from CPU as with not enough CPU the container will slow down while with missing memory it will be killed. That's why requirements and limits should be set to same value.

The change should happen in check template definition
https://github.com/stackrox/kube-linter/blob/24d1fe33226db4ffd03acadda46198a1338c4025/pkg/builtinchecks/yamls/unset-memory-requirements.yaml#L11
As well as in the related tests: 
https://github.com/stackrox/kube-linter/blob/e68f7d1cff4bf550c6fbadbc6a6509007b3c1e53/e2etests/bats-tests.sh#L916

- #694 
- [What if you specify a container's limit, but not its request?](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/#what-if-you-specify-a-container-s-limit-but-not-its-request)

PR #734, opened 2024-02-26
Update unset-memory-requirements and unset-cpu-requirements

### LINKED ISSUE #694: Change unset-cpu-requirements check
Currently the `unset-cpu-requirements` requires requests and limits while in general having only requests should be sufficient. Requests are essential for scheduling, and since omitting limits would simply allow the application to utilize a slightly higher amount of CPU if necessary, we should no longer recommend specifying both requests and limits.

The change should happen in check definition

https://github.com/stackrox/kube-linter/blob/e68f7d1cff4bf550c6fbadbc6a6509007b3c1e53/pkg/builtinchecks/yamls/unset-cpu-requirements.yaml#L11

As well in related tests: 
https://github.com/stackrox/kube-linter/blob/e68f7d1cff4bf550c6fbadbc6a6509007b3c1e53/e2etests/bats-tests.sh#L895
---
- #695

bgalek:
Hi! Happy to fix #694 #695!

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,11 +5,11 @@
 
 # What is KubeLinter?
 
-KubeLinter analyzes Kubernetes YAML files and Helm charts, and checks them against a variety of best practices, with a focus on production readiness and security. 
+KubeLinter analyzes Kubernetes YAML files and Helm charts, and checks them against a variety of best practices, with a focus on production readiness and security.
 
 KubeLinter runs sensible default checks, designed to give you useful information about your Kubernetes YAML files and Helm charts. This is to help teams check early and often for security misconfigurations and DevOps best practices. Some common examples of these include running containers as a non-root user, enforcing least privilege, and storing sensitive information only in secrets.
 
-KubeLinter is configurable, so you can enable and disable checks, as well as create your own custom checks, depending on the policies you want to follow within your organization. 
+KubeLinter is configurable, so you can enable and disable checks, as well as create your own custom checks, depending on the policies you want to follow within your organization.
 
 When a lint check fails, KubeLinter reports recommendations for how to resolve any potential issues and returns a non-zero exit code.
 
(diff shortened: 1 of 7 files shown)


README.md
janisz, line 1: I like the idea of splitting long lines but please do it in a separated PR. Ideally we should have a CI check to prevent long lines in the future.

pkg/builtinchecks/yamls/unset-memory-requirements.yaml
```diff
@@ -1,13 +1,13 @@
...
-  Set memory requests and limits for your container based on its requirements.
+  Set memory limits for your container based on its requirements.
   Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.
 scope:
   objectKinds:
     - DeploymentLike
 template: "memory-requirements"
 params:
-  requirementsType: "any"
+  requirementsType: "request"
```
janisz, line 11: As mentioned in issue we need limits for memory as request if not set will be copied from limits.
- https://github.com/stackrox/kube-linter/issues/695
```suggestion
  requirementsType: "limits"
```

pkg/builtinchecks/yamls/unset-cpu-requirements.yaml
```diff
@@ -4,10 +4,10 @@ scope:
   objectKinds:
     - DeploymentLike
 remediation: >-
-  Set CPU requests and limits for your container based on its requirements.
+  Set CPU limits for your container based on its requirements.
```
janisz, line 7: As mentioned in the issue we need requests as it's required for scheduling and with no limits we allow bursts in CPU usage
- https://github.com/stackrox/kube-linter/issues/694
```suggestion
  Set CPU requests for your container based on its requirements.
```

janisz (changes requested):

janisz (approved):

janisz:
@bgalek Thanks! 🎉

Issue #748, opened 2024-03-17
[FEATURE_REQUEST] Expand `run-as-non-root` template to verify runAsGroup field is nonzero.


ariyonaty:
**Description of the problem/feature request**
In addition to it being a best security practice for pods to have `runAsUser` set to a non-zero value, it is also recommended that the `GID`, determined by either the runtime default security context or the `runAsGroup` field, is set to a non-zero value.

Would like to propose either creating a new template/check or extend the existing `run-as-non-root` template to check against the `runAsGroup` field.

**Description of the existing behavior vs. expected behavior**
Below is a snippet of behavior when `runAsUser` set to `0`. Expected behavior would be along similar lines.
```
$ ./kube-linter lint ~/Documents/kube-linter/pkg/command/lint/testdata/valid-pod.yaml
KubeLinter 0.6.8

/home/user/Documents/kube-linter/pkg/command/lint/testdata/valid-pod.yaml: (object: <no namespace>/homebrew-demo /v1, Kind=Pod) container "homebrew-test" is not set to runAsNonRoot (check: run-as-non-root, remediation: Set runAsUser to a non-zero number and runAsNonRoot to true in your pod or container securityContext. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for details.)

Error: found 1 lint errors
```

**Additional context**
Not particularly familiar with Go, but would be glad to take a crack at this. 
Just let me know if there's any particular preference to either extend, or create new template, (or any other helpful suggestions/pointers).

janisz:
I think it could be added to https://docs.kubelinter.io/#/generated/checks?id=run-as-non-root

janisz:
@ariyonaty - CEL expression solution using the new template:

- https://github.com/stackrox/kube-linter/pull/1012

## Simple Solution
```yaml
customChecks:
  - name: run-as-non-root-group
    template: cel-expression
    params:
      check: |
        has(object.spec.containers) && 
        object.spec.containers.exists(c, 
          has(c.securityContext.runAsGroup) && 
          c.securityContext.runAsGroup == 0) ?
        "Container should not run as root group (runAsGroup: 0)" : ""
    scope:
      objectKinds: [DeploymentLike]
```

PR #804, opened 2024-07-08
Expand `run-as-non-root` template to verify `runAsGroup` field is nonzero.

ariyonaty:
Related issue: #748

This PR introduces changes which expand the `run-as-non-root` template/check to verify `runAsGroup` field is set to a non-zero value.

The `runAsGroup` field specifies the group ID under which the container’s process should run. 
If left unspecified, the [GID defaults to 0](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#:~:text=If%20this%20field%20is%20omitted%2C%20the%20primary%20group%20ID%20of%20the%20containers%20will%20be%20root(0)).

Snippet of build in action where `pod.yaml` for first execution initially didn't set `runAsGroup`, second execution set the field to 0, and finally set to a non-zero value:
image


janisz (commented):
could you add unit tests for new code?

pkg/templates/runasnonroot/template.go
```diff
@@ -53,10 +63,25 @@ func init() {
 				var results []diagnostic.Diagnostic
 				for _, container := range podSpec.AllContainers() {
 					runAsUser := effectiveRunAsUser(podSpec.SecurityContext, container.SecurityContext)
-					// runAsUser explicitly set to non-root. All good.
-					if runAsUser != nil && *runAsUser > 0 {
+					runAsGroup := effectiveRunAsGroup(podSpec.SecurityContext, container.SecurityContext)
+					// runAsUser and runAsGroup explicitly set to non-root. All good.
+					if (runAsUser != nil && *runAsUser > 0) && (runAsGroup != nil && *runAsGroup > 0) {
```
janisz, line 68: How about extracting a function to check for non zero int pointer? 
```
func isNonZero(number *int64) bool {
	return number != nil && *number > 0
}
```

pkg/templates/runasnonroot/template.go
```diff
@@ -14,6 +14,8 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+const templateKey = "run-as-non-root"
```
janisz, line 17: I think we need to keep `-user` as this will be a breaking change
```suggestion
const templateKey = "run-as-non-root-user"
```

PR #816, opened 2024-07-18
chore: move tools to a separated module

janisz:
External tools pollute `go.mod` and make it harder to update project. In this PR new module is created for golangci-lint and goreleaser and Go is updated to 1.22.5


Makefile
```diff
@@ -1,5 +1,7 @@
-.PHONY: deps
```
janisz, line 1: This was a mistake we create a file here

tool-imports/tools.go
```diff
@@ -10,5 +10,4 @@ package toolimports
 import (
 	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
 	_ "github.com/goreleaser/goreleaser"
-	_ "k8s.io/cli-runtime/pkg/resource" // This is imported because we want to be able to require it at a certain version, since otherwise Helm breaks.
```
janisz, line 13: let's see if we really need it

stehessel (approved):

PR #866, opened 2024-11-16
Generate release tarballs with vendored modules with .goreleaser.yaml

mattmcadoo:
Used for distros/circumstances where pulling external dependencies directly during build is not allowed.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -124,3 +124,8 @@ docker_signs:
 
 changelog:
   use: github-native
+
+source:
+  enabled: true
+  files:
+    - vendor


CODE_OF_CONDUCT.md
rhybrillou, line 1: Please restore the contents of this file.

rhybrillou (changes requested):

CODE_OF_CONDUCT.md
mattmcadoo: That wasn't supposed to be pulled in on the PR, I was trying something and used that file as the test...

mattmcadoo: Bogus commit removed.  I didn't realize that an open PR would pull commits in from my fork of the repo after the PR was generated.

PR #875, opened 2024-12-02
chore: use only native caches from actions/setup-go and golangci/golangci-lint-action

mmorel-35:
Those two steps are not of use anymore as caches are alredy handle by actions

* actions/setup-go@v5 
  * ~/.cache/go-build
  * ~/go/pkg/mod
* golangci/golangci-lint-action@v6
  * ~/.cache/golangci-lint

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,18 +29,6 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - name: Go Build Cache
-        uses: actions/cache@v4
-        with:
-          path: ~/.cache
-          key: ${{ runner.os }}-go-build-${{ hashFiles('**/go.sum') }}
-
-      - name: Go Mod Cache
-        uses: actions/cache@v4
-        with:
-          path: ~/go/pkg/mod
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -51,7 +39,10 @@ jobs:
           path: bin
 
       - name: Run lint checks
-        run: make lint
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: v1.59.1
```
janisz, line 44: Currently we are using higher version
https://github.com/stackrox/kube-linter/blob/0c713814a586e4f587f1f04335bfe28ce0847265/tool-imports/go.mod#L6

janisz (commented):
This change is reasonable, but we've previously pinned the linter in a separate `go.mod` file to allow Dependabot to handle automatic updates. Pinning it in the YAML file instead prevents these updates. 😞 

We also need to ensure that `setup-go`'s cache handles the linter properly by including its `go.sum` file in the cache configuration (as described [here](https://github.com/actions/setup-go#caching-dependency-files-and-build-outputs)) and ensuring the linter cache is stored correctly.

Issue #609, opened 2023-08-04
[FEATURE_REQUEST] Support Restart Policy


BlythMeister:
**Description of the problem/feature request**
Add support to check the `restartPolicy` is set to `Always`

**Additional context**
We are migrating from datree and this was a check they have: https://hub.datree.io/built-in-rules/ensure-valid-restart-policy

PR #915, opened 2025-02-21
feature(lint): Adds a new template linter to check the restart policy for all deployments like objects

jonathanmdr:
Closes: #609 

This PR introduces changes to include a new checker to validate the restart policy for all deployments like objects.


jonathanmdr:
Hi @rhybrillou and @janisz!

Can you spend some time to review this PR?

If there's any feedback about this approach, I will be happy to follow up on it 😄

e2etests/bats-tests.sh
```diff
@@ -785,6 +785,30 @@ get_value_from() {
   [[ "${count}" == "2" ]]
 }
 
+@test "restart-policy" {
+  tmp="tests/checks/restart-policy.yaml"
+  cmd="${KUBE_LINTER_BIN} lint --include restart-policy --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}
+
+  message1=$(get_value_from "${lines[0]}" '.Reports[0].Object.K8sObject.GroupVersionKind.Kind + ": " + .Reports[0].Diagnostic.Message')
```
rhybrillou, line 793: IMHO, it would be good to add the object name to the messages in order to ease debugging.
As far as I can see, there are more objects in the test file than there are messages/reports here. While it is easy to deduce which ones should be present in the report, I would prefer to have the debugging information state more explicitly what objects are expected to be reported by the linter.

docs/generated/checks.md
```diff
@@ -531,6 +531,15 @@ value: '[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+'
 ```yaml
 key: owner
 ```
+## restart-policy
+
+**Enabled by default**: No
+
+**Description**: Indicates when a object deployment's like doesn't use a restart policy
```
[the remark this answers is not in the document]
jonathanmdr: Adjusted

pkg/builtinchecks/yamls/restart-policy.yaml
```diff
@@ -0,0 +1,8 @@
+name: "restart-policy"
+description: "Indicates when a object deployment's like doesn't use a restart policy"
```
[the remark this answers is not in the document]
jonathanmdr: Adjusted

e2etests/bats-tests.sh
jonathanmdr: Adjusted

rhybrillou: I think the valid object in the test file did belong there too.
What I was thinking for the test was something along the lines (if the report objects expose the required fields):
```
Deployment fire-deployment-never: object has a restart policy defined with 'Never' but the only accepted restart policies are '[Always OnFailure]'
```

rhybrillou (approved):

janisz (approved):

pkg/templates/restartpolicy/template_test.go
```diff
@@ -0,0 +1,118 @@
...
+	s.addDeploymentWithoutRestartPolicy(withoutRestartPolicy)
+	s.addDeploymentWithEmptyRestartPolicy(emptyRestartPolicy)
+	s.addDeploymentWithRestartPolicy(restartPolicyNever, coreV1.RestartPolicyNever)
+
+	s.Validate(s.ctx, []templates.TestCase{
+		{
+			Param: params.Params{},
+			Diagnostics: map[string][]diagnostic.Diagnostic{
+				withoutRestartPolicy: {
+					{Message: "object has a restart policy defined with '' but the only accepted restart policies are '[Always OnFailure]'"},
```
mloiseleur, line 68: Since the default is _Always_, I'm not sure to understand why it's expected to fail when restartPolicy is not defined.

mloiseleur: On datree documentation, it says that empty value is correct:
https://hub.datree.io/built-in-rules/ensure-valid-restart-policy

jonathanmdr: For the `kube-linter`, the rule is not the same as `datree`; the values accepted by `kube-linter` are `Always` and `OnFailure`.
Any other value is considered invalid, as `Never` or an empty value.

PR #922, opened 2025-03-11
Add a New Check For Annotations on PersistentVolumeClaims

hanife-Orthopy:


janisz:
LGTM
but please add e2e test https://github.com/stackrox/kube-linter/blob/main/e2etests/bats-tests.sh

Issue #932, opened 2025-03-19
Why ignore objects without pods in the dangling-hpa check?


mboutet:
_Originally posted as a discussion in https://github.com/stackrox/kube-linter/discussions/931but but reopening as an issue as it appears that discussions rarely get answered._

----

https://github.com/stackrox/kube-linter/blob/51eb58ee7d78cdf48c5b425774e98573aab11ba4/pkg/templates/danglinghpa/template.go#L81-L84

I have a HPA targeting a `KafkaNodePool` resource but because of that condition in the code, it doesn't go further and it flags my HPA as being dangling.

Why is it needed to check if the potential target object has pods? Why is it not sufficient for this check to only make sure that the `scaleTargetRef` exists (as it's already doing later in the loop as shown below)?

https://github.com/stackrox/kube-linter/blob/51eb58ee7d78cdf48c5b425774e98573aab11ba4/pkg/templates/danglinghpa/template.go#L86-L91

Thank you

mboutet:
Well, never mind, the issue is actually that `kube-linter` [doesn't parse all custom objects](https://github.com/stackrox/kube-linter/issues/24) and my issues was with a `KafkaNodePool` custom resource targeted by an HPA.

Issue #24, opened 2020-10-29
Support CRDs


viswajithiii:
We should support linting custom resources. Some suggestions we've [got](https://www.reddit.com/r/kubernetes/comments/jjpqn5/introducing_kubelinter_an_open_source_linter_for/gaf4gvg) from users are: Prometheus Operator (specifically, PrometheusRule manifests), Cert Manager, ArgoCD, and Istio.

erNail:
Hey @viswajithiii , is there any update regarding this issue?

flowdopip:
Hey, any update on this?

awalther52:
Any update on this?

viswajithiii:
I am no longer a maintainer, but tagging current maintainers @janisz @dhaus67

jhoover4:
Wanted to check on this as well. Is there any way I could help? Specifically I'm familiar with Argo and Istio resources

CyDickey-msr:
This would be a nice feature. Similar to how [yannh/kubeconform](https://github.com/yannh/kubeconform?tab=readme-ov-file#CustomResourceDefinition-CRD-Support) supports CRD schemas from https://github.com/datreeio/CRDs-catalog.

greg-swift-yohana:
At least for Istio can theoretically for moving forward can just target stackrox/kube-linter#555

PR #933, opened 2025-03-19
Do not ignore objects without pod template in `dangling-horizontalpodautoscaler`

mboutet:
Fixes https://github.com/stackrox/kube-linter/issues/932.

I personally don't understand why the object referenced in the `scaleTargetRef` of the HPA needs to have a pod template in its manifest. I find this pre-condition to be out of scope for this check. The check should only ensure that the referenced object in `scaleTargetRef` exists in the linting context. There are way too many custom resources that have the `/scale` sub-resource without defining by default a pod template in their manifest.

I also checked the original PR https://github.com/stackrox/kube-linter/pull/271 implementing this check and I could not find any comment or commits explaining why this pre-condition is necessary.

### DIFF
--- a/pkg/templates/danglinghpa/template.go
+++ b/pkg/templates/danglinghpa/template.go
@@ -6,7 +6,6 @@ import (
 	"golang.stackrox.io/kube-linter/pkg/check"
 	"golang.stackrox.io/kube-linter/pkg/config"
 	"golang.stackrox.io/kube-linter/pkg/diagnostic"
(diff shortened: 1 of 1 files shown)


mboutet:
Well, never mind, the issue is actually that `kube-linter` [doesn't parse all custom objects](https://github.com/stackrox/kube-linter/issues/24) and my issues was with a `KafkaNodePool` custom resource targeted by an HPA.

Issue #899, opened 2025-01-24
[FEATURE_REQUEST] Publish the config file schema on schemastore.org


tarampampam:
**Description of the problem/feature request**

It would be great if someone could describe the config file format using JSON Schema and publish it on [schemastore.org](https://www.schemastore.org/json/), which has broad support across a variety of IDEs.

**Additional context**

This would allow for real-time configuration validation and suggestions for all possible configuration options, greatly improving the user experience.

janisz:
Fortunately our config is quite simple and we need to add a schema for `.kube-linter.yaml` and `.kube-linter.yml` files.
I think we should keep the schema also in the repo. There is a question if we should make linters names part of schema or not.

PR #978, opened 2025-07-03
Add a basic schema validation for config files

kquinsland:
Good for basic validation.
More work (and probably some CI automation?) would be needed for precisely validating each of the possible custom checks.

This should resolve: https://github.com/stackrox/kube-linter/issues/899

### DIFF
--- a/schemas/kube-lint-config.json
+++ b/schemas/kube-lint-config.json
@@ -1,0 +1,89 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "KubeLinter Configuration",
+    "type": "object",
+    "properties": {
+        "checks": {
+            "type": "object",
+            "description": "Configure which built-in checks to run or skip",
+            "properties": {
+                "doNotAutoAddDefaults": {
+                    "type": "boolean",
+                    "description": "Disable all built-in checks"
+                },
+                "addAllBuiltIn": {
+                    "type": "boolean",
+                    "description": "Enable all built-in checks. Takes precedence over doNotAutoAddDefaults"
+                },
+                "ignorePaths": {
+                    "type": "array",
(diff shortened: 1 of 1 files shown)


schemas/kube-lint-config.json
```diff
@@ -0,0 +1,89 @@
...
+            "type": "object",
+            "description": "Configure which built-in checks to run or skip",
+            "properties": {
+                "doNotAutoAddDefaults": {
+                    "type": "boolean",
+                    "description": "Disable all built-in checks"
+                },
+                "addAllBuiltIn": {
+                    "type": "boolean",
+                    "description": "Enable all built-in checks. Takes precedence over doNotAutoAddDefaults"
```
rhybrillou, line 16:
```suggestion
                    "description": "Enable all built-in checks. Users can explicitly opt-out checks using exclude. Takes precedence over doNotAutoAddDefaults"
```

schemas/kube-lint-config.json
```diff
@@ -0,0 +1,89 @@
...
+                        "type": "string"
+                    }
+                }
+            },
+            "additionalProperties": false
+        },
+        "customChecks": {
+            "type": "array",
+            "description": "User-defined checks based on templates",
+            "items": {
```
rhybrillou, line 45: If I read the file at `https://github.com/stackrox/kube-linter/blob/main/pkg/config/check.go` correctly, there is one more parameter for the `customeChecks` array: `description`. Would you mind adding it ?

rhybrillou (commented):
Hello @kquinsland 
Thanks a lot for your contribution.
Can you please cross-check the properties of the `customChecks` array object ?

rhybrillou (changes requested):
Could you check again the parameters for the `customChecks` array items ? I think a `description` property is missing there.

rhybrillou (approved):

Issue #705, opened 2024-01-29
[FEATURE_REQUEST]


josegomezr:
**Description of the problem/feature request**

A check to detect containers referring to configmaps/secret keys that do not exist in the current release via `valueFrom` (`v1.EnvVarSource`)

**Description of the existing behavior vs. expected behavior**

In some of our workloads we have containers with environment variables defined as:

```
# [...] clipped for brevity
- name: AIRBRAKE_PROJECT_KEY
  valueFrom:
    secretKeyRef:
      name: secrets
      key: airbrake-project-key
# [...] clipped for brevity
```

But it has happened a couple of times that a typo in the `key` slips and we notice it at the very last moment when the helm release fails to complete.

**Additional context**

The minimum reproduceable manifest could be something like:

```
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-deployment
  namespace: example
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: example
  template:
    metadata:
      labels:
        app.kubernetes.io/name: example
    spec:
      containers:
      - name: example-container
        image: internal.container.registry/an-image:v1
        env:
          - name: MY_DANGLING_ENV
            valueFrom:
              secretKeyRef:
                name: existing-secret
                key: with-a-typo # <-- notice here
---
apiVersion: v1
kind: Secret
metadata:
  name: example-deployment
  namespace: example
type: Opaque
data:
  without-a-typo: "YmFzZTY0Li4uCg=="
```

And the output should indicate which deployment or the container (whatever is more comfy) is referring to an unknown secret/config-map.

Currently there's no output of this, I do have a draft in a fork I can contribute to. I'm missing some tests but it sort-of works (golang is not my _native speak_):

https://github.com/stackrox/kube-linter/compare/main...josegomezr:kube-linter:env_source_check

Nexusrex18:
is this feature still needed ? I'd like to work on it .

Nexusrex18:
I have implemented a draft solution in which i implemented the `env-value-from` check from #705, based on @josegomezr WIP branch. It flags non-existent secrets/ConfigMaps in `valueFrom` fields for `DeploymentLike` objects. but I was asking that adding `env-value-from` to `defaultchecks.List` appropriate, or should it be opt-in via `config.yaml`?

PR #1032, opened 2025-10-06
feat: add env-value-from as opt-in check for #705

Nexusrex18:
Implements env-value-from check from #705  as opt-in (via config.yaml 'include'). If maintainers prefer, can add to default checks in a follow-up.


janisz (changes requested):
Looks good, the only problem is duplicated code

tests/checks/env-var-value-from.yml
```diff
@@ -0,0 +1,84 @@
+---
+# This should fail due to key mismatch (chimpmunk vs chimpmun)
```
janisz: Ideally we should have test case for config map too

janisz:
Nice! Thank you for your contribution!

I plan to release once get https://github.com/stackrox/kube-linter/pull/1045 merged

Issue #606, opened 2023-08-02
[BUG] 


PascalTurbo:
**System info:**
 - OS: MacOS, also Docker Image

**Describe the bug**
kube-linter returns an error (see below) but the HelmChart is valid and installable

**To Reproduce**
Checkout this code https://github.com/PascalTurbo/stackrox-kube-linter-bug-example and run `kube-linter lint -v chart`

**Output**
```
kube-linter lint -v chart

Warning: failed to load object from chart/templates/pipeline.yaml: failed to decode: no kind "Pipeline" is registered for version "tekton.dev/v1beta1" in scheme "pkg/runtime/scheme.go:100"
Warning: no valid objects found.
```

janisz:
We do not registered tekton schema. Probably we should do that. We have an issue about this:

- #24 

Below patch solves this issue but it adds new dependency and that will be hard to maintain. On the other hand we could add schemas for mature project that are part of CNCF or some other org (e.g. we already have exception for OpenShift) 

```diff
diff --git a/go.mod b/go.mod
index 3d2db6e..4d36d51 100644
--- a/go.mod
+++ b/go.mod
@@ -18,17 +18,20 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.16.0
 	github.com/stretchr/testify v1.8.4
+	github.com/tektoncd/pipeline v0.40.0
 	helm.sh/helm/v3 v3.12.2
 	k8s.io/api v0.27.4
 	k8s.io/apimachinery v0.27.4
 	k8s.io/cli-runtime v0.27.4
 	k8s.io/client-go v0.27.4
-	k8s.io/gengo v0.0.0-20220902162205-c0856e24416d
+	k8s.io/gengo v0.0.0-20221011193443-fad74ee6edd9
 )
 
diff --git a/pkg/lintcontext/parse_yaml.go b/pkg/lintcontext/parse_yaml.go
index 6a6b6f3..ce5b84e 100644
--- a/pkg/lintcontext/parse_yaml.go
+++ b/pkg/lintcontext/parse_yaml.go
@@ -13,6 +13,7 @@ import (
 	y "github.com/ghodss/yaml"
 	ocsAppsV1 "github.com/openshift/api/apps/v1"
 	"github.com/pkg/errors"
+	tekton "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"golang.stackrox.io/kube-linter/pkg/k8sutil"
[...]

PR #1033, opened 2025-10-08
Add kubeconform template for schema validation

janisz:
This PR introduces a new kubeconform template that validates k8s resources against their schemas using the [kubeconform](https://github.com/yannh/kubeconform) library.

Fixes:
- #36 

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #1012, opened 2025-08-27
Add CEL template

janisz:
## Add CEL Template for Custom Expression-Based Checks

This PR introduces a new `cel-expression` template that enables users to write custom checks using the [Common Expression Language (CEL)](https://github.com/google/cel-spec). This powerful addition allows for flexible, expression-based validation of Kubernetes resources without requiring Go code changes.

### 🚀 Features

- **CEL Expression Template**: New `cel-expression` template in `pkg/templates/cel/template.go:18`
- **Flexible Resource Validation**: Supports any Kubernetes resource type through `objectkinds.Any`
- **Expression-Based Logic**: Write custom validation logic using [CEL syntax](https://github.com/google/cel-spec/blob/master/doc/langdef.md)
- **Access to Full Context**: CEL expressions have access to both the subject resource and all other objects in the cluster

### 🔧 How It Works

The template provides two variables to CEL expressions:
- `subject`: The current Kubernetes resource being evaluated
- `objects`: Array of all Kubernetes resources in the current context

CEL expressions should return:
- An empty string `""` if the check passes
- A descriptive string message if the check fails

### 📝 Usage Examples

Replace complex template logic with simple CEL expressions:

**Example 1: Forbidden Annotation Check**
```yaml
customChecks:
  - name: "cel-forbidden-annotation"
    description: "Detect forbidden reloader annotation"
    template: "cel-expression"
    scope:
      objectKinds:
        - DeploymentLike
    params:
      check: |
        has(subject.metadata.annotations) && 
        subject.metadata.annotations["reloader.stakater.com/auto"] == "true" ? 
        "Object has reloader annotation" : ""
```

**Example 2: IRSA Role Validation**
```yaml
customChecks:
  - name: "invalid-irsa-role"
    description: "Validate EKS IRSA role ARN format"
    template: "cel-expression"
    scope:
      objectKinds:
        - ServiceAccount
    params:
      check: |
        subject.metadata.annotations["eks.amazonaws.com/role-arn"].matches("^arn:aws:iam::\\d{12}:role/[\\w+=,.@-]{1,64}$") ? 
        "" : "Invalid EKS IAM role ARN format"
```

### 🧪 Testing

- **E2E Tests**: New test case in `e2etests/bats-tests.sh` demonstrates the CEL template replacing traditional forbidden annotation checks
- **Unit Tests**: Comprehensive test coverage in `pkg/templates/cel/template_test.go`
[...]

Issue #848, opened 2024-10-08
[FEATURE_REQUEST] Argo Rollouts support


romancin:
**Description of the problem/feature request**
Currently, we are starting to migrate our applications from Kubernetes `Deployments` to Argo `Rollouts`, which provides advanced deployment strategies, like bluegreen or canary. As stated in the [Argo Rollouts docs](https://argo-rollouts.readthedocs.io/en/stable/concepts/#rollout):

> A Rollout is Kubernetes workload resource which is equivalent to a Kubernetes Deployment object.
> It is intended to replace a Deployment object in scenarios when more advanced deployment or progressive 
> delivery functionality is needed

**Description of the existing behavior vs. expected behavior**
When deleting our deployment, the `dangling-service` check included in kube-linter starts failing, because it can't match any pods, because in this case, the pods are now controlled by Argo Rollouts.

Input file:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    run: my-app
  name: my-app
  namespace: default
spec:
  ports:
  - name: http
    nodePort: 30395
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    run: my-app
  type: NodePort
---
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  annotations:
    ignore-check.kube-linter.io/no-read-only-root-fs: Pending to validate if if works with read-only root filesystem
    ignore-check.kube-linter.io/run-as-non-root: Pending to validate if the pod works with other user than root
    owner: platform
  name: my-app
  namespace: default
spec:
  minReadySeconds: 3
  replicas: 2
  selector:
    matchLabels:
      run: my-app
  strategy:
    canary:
      steps:
      - setWeight: 50
      - pause:
          duration: 10s
  template:
    metadata:
      annotations:
        owner: platform
      labels:
        run: my-app
    spec:
      containers:
      - image: docker.io/my-app:1.0.0
        imagePullPolicy: Always
        name: my-app
        ports:
        - containerPort: 80
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          httpGet:
            httpHeaders:
            - name: Host
              value: health
            path: /
            port: 80
          initialDelaySeconds: 5
          periodSeconds: 1
          timeoutSeconds: 1
        resources:
          limits:
            cpu: "1.5"
            memory: 800Mi
          requests:
            cpu: "0.3"
            memory: 200Mi
---
apiVersion: autoscaling/v2
[...]

PR #1037, opened 2025-10-08
Add fallback decoder for unknown resources to handle CRDs

janisz:
This PR adds a fallback when decoding unknown resources to be able to handle CRDs. The schema validation could be then performed with kubeconform check from https://github.com/stackrox/kube-linter/pull/1033 and CEL from https://github.com/stackrox/kube-linter/pull/1012. This should fix https://github.com/stackrox/kube-linter/issues/606
It also deprecates the ` --fail-on-invalid-resource` as it could be replaced by a new `schema-validation` check or `kubeconform`
This could be considered as breaking change as it changes the semantic of ` --fail-on-invalid-resource` to fail on invalid yamls instead of failing on unknown objects or not matching schema but this could be mitigated by kubeconform.

Changes:
- Modified parseObjects to use unstructured decoder as fallback for unknown resource types
- Added comprehensive test suite covering standard K8s resources and CRDs
- Maintained backward compatibility for existing decode error handling
- Added test cases for Tekton Task CRD and other custom resources

The fallback allows kube-linter to parse CRDs like Tekton Pipelines without
failing, while delegating proper schema validation to specialized templates
like kubeconform and CEL expressions.

Fixes:
- #606
- #24 
- #208 

Refs:
- #848 

🤖 Generated with [Claude Code](https://claude.ai/code)

  Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/pkg/command/lint/command_test.go
+++ b/pkg/command/lint/command_test.go
@@ -16,7 +16,6 @@ func TestCommand_InvalidResources(t *testing.T) {
 		failure bool
 		output  string
 	}{
-		{name: "InvalidPodResource", cmd: createLintCommand("./testdata/invalid-pod-resources.yaml", "--fail-on-invalid-resource"), failure: true},
 		{name: "InvalidYAML", cmd: createLintCommand("./testdata/invalid.yaml", "--fail-on-invalid-resource"), failure: true},
 		{name: "NonexistentFile", cmd: createLintCommand("./testdata/foo-bar.yaml", "--fail-on-invalid-resource"), failure: true},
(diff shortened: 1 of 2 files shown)


pkg/templates/kubeconform/template.go
```diff
@@ -33,6 +34,13 @@ func init() {
 }
 
 func validate(p params.Params) (check.Func, error) {
+	// Create cache directory if it doesn't exist
+	if p.Cache != "" {
+		if err := os.MkdirAll(p.Cache, 0755); err != nil {
```
rhybrillou: Please address the gosec error (drop directory permissions for other)

rhybrillou (approved):

PR #1189, opened 2026-05-21
fix: extract replicationcontroller selectors

### LINKED ISSUE #1188: [BUG] mismatching-selector registers ReplicationController but appears to skip it
## **System info:**
 - OS: macOS

## **Describe the bug**
The `mismatching-selector` check appears to advertise `ReplicationController` support through its `DeploymentLike` scope, but `ReplicationController` objects are skipped during selector extraction.

`DeploymentLike` includes `ReplicationController`, but `mismatching-selector` calls `extract.Selector(object.K8sObject)`. For `ReplicationController`, `spec.selector` is a plain `map[string]string`, while `extract.Selector` only handles special-cased `DeploymentConfig` / `CronJob` selectors or reflected `Spec.Selector` fields typed as `*metav1.LabelSelector`.

As a result, a `ReplicationController` with mismatched `spec.selector` and pod template labels is not reported.

## **To Reproduce**

1. Save the sample YAML below as `rc-mismatching-selector.yaml`.
2. Run:
   ```bash
   kube-linter lint --include mismatching-selector --do-not-auto-add-defaults rc-mismatching-selector.yaml
   ```
3. Observe that no `mismatching-selector` finding is reported for the `ReplicationController`.

## **Sample YAML input**
```yaml
apiVersion: v1
kind: ReplicationController
metadata:
  name: rc-selector-mismatch
spec:
  selector:
    app: api
    tier: backend
  template:
    metadata:
      labels:
        app: api
        tier: frontend
    spec:
      containers:
      - name: app
        image: nginx
```

## **Expected behavior**
The `mismatching-selector` check should report that the `ReplicationController` selector does not match the pod template labels.

Either `ReplicationController` should be removed from the effective `DeploymentLike` scope for this check, or `extract.Selector` should convert `ReplicationController.spec.selector` into a `metav1.LabelSelector{MatchLabels: ...}` similar to how `DeploymentConfig` is handled.

## **Screenshots**
Not applicable.

## **Additional context**
The check is registered with `scope.objectKinds: DeploymentLike`, and `DeploymentLike` includes `ReplicationController`.

The issue seems to come from selector extraction rather than object-kind matching: `extract.Selector` does not currently handle selector maps from `ReplicationController`.

GeneCodeSavvy:
Resolves #1188

### DIFF
--- a/pkg/extract/pod_spec_test.go
+++ b/pkg/extract/pod_spec_test.go
@@ -1,0 +1,33 @@
+package extract
+
+import (
+	"reflect"
+	"testing"
+
+	coreV1 "k8s.io/api/core/v1"
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestSelectorExtractsReplicationControllerSelector(t *testing.T) {
+	rc := &coreV1.ReplicationController{
+		Spec: coreV1.ReplicationControllerSpec{
+			Selector: map[string]string{
+				"app": "web",
+			},
+		},
+	}
+
+	selector, found := Selector(rc)
+	if !found {
+		t.Fatal("expected selector to be found")
+	}
+
+	expected := &metaV1.LabelSelector{
+		MatchLabels: map[string]string{
+			"app": "web",
+		},
+	}
+	if !reflect.DeepEqual(selector, expected) {
+		t.Fatalf("expected selector %#v, got %#v", expected, selector)
+	}
+}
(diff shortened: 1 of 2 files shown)


rhybrillou (approved):

pkg/extract/pod_spec_test.go
```diff
@@ -0,0 +1,33 @@
+package extract
+
+import (
+	"reflect"
+	"testing"
+
+	coreV1 "k8s.io/api/core/v1"
```
rhybrillou, line 7:
```suggestion

	"github.com/stretchr/testify/require"
	coreV1 "k8s.io/api/core/v1"
```

pkg/extract/pod_spec_test.go
```diff
@@ -0,0 +1,33 @@
...
+			Selector: map[string]string{
+				"app": "web",
+			},
+		},
+	}
+
+	selector, found := Selector(rc)
+	if !found {
+		t.Fatal("expected selector to be found")
+	}
```
rhybrillou, line 23:
```suggestion
	require.True(t, found, "expected selector to be found")
```

rhybrillou (commented):
The change looks good overall.
Could you please use testify packages (assert or require) for the test validation instead of manual checks and calls to `t.Fatal` ?

PR #1197, opened 2026-06-05
Accept probe port exposed via container args or command

arpitjain099:
Fixes #1086

## Problem

The `liveness-port`, `readiness-port` and `startup-port` checks resolve the probe port only against the container's declared `containerPort` list. A `containerPort` entry is informational: a process can listen on any port regardless of whether it is declared. When the probe port is wired up through a flag instead of a declared port, the checks report a false positive for a port that is actually being served.

The opentelemetry-operator chart (0.100.0) hits this. The manager container passes its health-probe address purely through args and only declares the metrics port:

```yaml
containers:
  - name: manager
    args:
      - --metrics-addr=0.0.0.0:8080
      - --health-probe-addr=:8081
    ports:
      - containerPort: 8080
        name: metrics
        protocol: TCP
    livenessProbe:
      httpGet:
        path: /healthz
        port: 8081
```

## Before

`8081` is not in the declared ports, so the check fires even though the process listens on it:

```
container "manager" does not expose port 8081 for the HTTPGet
(check: liveness-port)
```

## After

`CheckProbePort` now also looks for a numeric probe port as a standalone integer token in the container `args` and `command` before flagging it. With the change above the container no longer fires, which matches the expected behavior in the issue.

Because all three checks share `CheckProbePort`, the fix covers `liveness-port`, `readiness-port` and `startup-port` at once.

## Scope and limitations

- The match is intentionally conservative. Only a numeric probe port is matched against args/command; a named (string) port still has to resolve against a declared `containerPort` exactly as before.
- The port number is matched as a standalone integer token, so searching for `8081` does not match `18081`, `80818`, or a port that only appears as a substring of an image tag or version. Those cases still fire.
- This is a textual presence check on the args, not a parse of every CLI flag convention. It accepts the common case where the port number appears in the args (for example `--health-probe-addr=:8081`, `--addr=0.0.0.0:8081`, `--port 8081`). It does not attempt to confirm which flag the port belongs to.

## Tests

- `pkg/templates/util/check_probe_port_test.go` (new): direct unit tests for `CheckProbePort` and `argContainsPort`, including the issue #1086 args case, the command case, the string-port no-match, and the substring regression guards (`18081`, `80818` still fire).
- `pkg/templates/livenessport/template_test.go`: added a `PortExposedViaArgs` case to the existing suite.
- `tests/checks/liveness-port.yml`: added a `dont-fire-deployment-port-in-args` container reproducing the chart scenario end to end. The existing four fire-cases and their assertions are unchanged.

`go build ./...`, `go vet`, and `go test ./pkg/templates/... ./pkg/lintcontext/...` all pass.

### DIFF
--- a/pkg/templates/util/check_probe_port.go
+++ b/pkg/templates/util/check_probe_port.go
@@ -2,6 +2,8 @@ package util
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
 
 	"golang.stackrox.io/kube-linter/pkg/diagnostic"
 	v1 "k8s.io/api/core/v1"
@@ -25,7 +27,7 @@ func CheckProbePort(container *v1.Container, probe *v1.Probe) []diagnostic.Diagn
 	}
 
 	if httpProbe := probe.HTTPGet; httpProbe != nil {
-		if _, ok := ports[httpProbe.Port]; !ok {
+		if _, ok := ports[httpProbe.Port]; !ok && !probePortInArgs(container, httpProbe.Port) {
 			return []diagnostic.Diagnostic{{
 				Message: fmt.Sprintf("container %q does not expose port %s for the HTTPGet", container.Name, httpProbe.Port.String()),
 			}}
@@ -33,7 +35,7 @@ func CheckProbePort(container *v1.Container, probe *v1.Probe) []diagnostic.Diagn
 	}
 
 	if tcpProbe := probe.TCPSocket; tcpProbe != nil {
-		if _, ok := ports[tcpProbe.Port]; !ok {
+		if _, ok := ports[tcpProbe.Port]; !ok && !probePortInArgs(container, tcpProbe.Port) {
 			return []diagnostic.Diagnostic{{
 				Message: fmt.Sprintf("container %q does not expose port %s for the TCPSocket", container.Name, tcpProbe.Port.String()),
 			}}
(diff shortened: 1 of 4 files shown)


pkg/templates/util/check_probe_port.go
```diff
@@ -25,27 +27,85 @@ func CheckProbePort(container *v1.Container, probe *v1.Probe) []diagnostic.Diagn
...
+// still has to resolve against a declared containerPort, so it is left to the
+// caller's normal lookup.
+func probePortInArgs(container *v1.Container, port intstr.IntOrString) bool {
+	if port.Type != intstr.Int {
+		return false
+	}
+	portNum := port.IntValue()
+	if portNum <= 0 {
+		return false
+	}
```
rhybrillou, line 73: nit: would it make sense to check whether the port value is in the uint16 range (1-65355) ?

rhybrillou (commented):
Thank you for this contribution (and sorry for the very late review).
Overall, the change looks good. I think there might be false negatives if some command or argument contains the probe port as a substring that does not aim at exposing the port. Nevertheless, the current implementation brings value to the tool.

Could you please add the following test case to TestArgContainsPort ?
```
// possible false positive induced by negative command line arguments
{"--disable-port-80", "80", true}
```

PR #1206, opened 2026-06-18
Avoid env-var-secret false positives for SECRETS_* names

### LINKED ISSUE #1198: [BUG] env-var-secret rule is triggered by `SECRETS_DIR`
**System info:**
 - OS: linux

**Describe the bug**

In our helm chart, we declare a env var called `SECRETS_DIR`. This triggers the env-var-secret rule despite not containing a secret at all ut actually implementing the recommended strategy to mount the secrets.

**To Reproduce**
Declare a env var called `SECRETS_DIR`

**Sample YAML input**

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whatever-service
  labels:
    app.kubernetes.io/managed-by: Helm
spec:
  replicas: 1
  template:
    spec:
      serviceAccountName: whatever-service
      initContainers:
        - name: db-migrations
        - command:
          - command
          env:
          - name: SECRETS_DIR  # <-
            value: /mnt/secrets
```

**Expected behavior**

I want that env var not to match the rule

**Screenshots**

Error:

```text
/chart/_pipeline_rendered_chart.yaml: (object: <no namespace>/whatever-service apps/v1, Kind=Deployment) environment variable SECRETS_DIR in container "db-migrations" found (check: env-var-secret, remediation: Do not use raw secrets in environment variables. Instead, either mount the secret as a file or use a secretKeyRef. Refer to https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets for details.)
```

**Additional context**

Unfortunately, we do not have a way to pass in the ignore, as the actual env vars in the above resource definition are rendered by helm from values.yaml :-(

arpitjain099:
Fixes #1198.

The built-in `env-var-secret` check used `(?i).*secret.*`, which also matched names like `SECRETS_DIR`.

This narrows the built-in pattern so plural `secrets` suffixes are not flagged, and extends the fixture with a `SECRETS_DIR` case to keep that behavior covered.

Verification:
- go test ./pkg/builtinchecks/...
- go run ./cmd/kube-linter lint --include env-var-secret --do-not-auto-add-defaults --format json tests/checks/env-var-secret.yml

### DIFF
--- a/docs/generated/checks.md
+++ b/docs/generated/checks.md
@@ -222,7 +222,7 @@ IgnoredSecrets: []
 **Parameters**:
 
 ```yaml
-name: (?i).*secret.*
+name: (?i).*secret([^s].*|$)
 value: .+
 ```
 ## exposed-services
--- a/pkg/builtinchecks/yamls/env-var-secret.yaml
+++ b/pkg/builtinchecks/yamls/env-var-secret.yaml
@@ -8,5 +8,5 @@ scope:
     - DeploymentLike
 template: "env-var"
 params:
-  name: "(?i).*secret.*"
+  name: "(?i).*secret([^s].*|$)"
   value: ".+"
--- a/tests/checks/env-var-secret.yml
+++ b/tests/checks/env-var-secret.yml
@@ -28,6 +28,19 @@ spec:
 apiVersion: apps/v1
 kind: Deployment
 metadata:
+  name: secrets-dir
+spec:
+  template:
+    spec:
+      containers:
+        - name: app
+          env:
+            - name: SECRETS_DIR
+              value: /mnt/secrets
+---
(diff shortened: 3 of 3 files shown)


rhybrillou (approved):

rhybrillou (approved):

manuel-lang:
@arpitjain099  @rhybrillou Thanks a lot for implementing the changes and reviewing the PR. Is there anything more that is needed to get the PR merged? Can't wait to have this change available :)

arpitjain099:
@rhybrillou @manuel-lang the approval here goes back to early July and the thread has gone quiet since. Is there anything outstanding on my side, or is this simply queued behind other work?

PR #1211, opened 2026-06-26
Add RBAC privilege escalation checks

sthadka:
## Summary
- Add 8 new security checks detecting well-known Kubernetes RBAC privilege escalation patterns
- All checks use the existing `cel-expression` template — zero Go code, pure YAML definitions
- Checks are based on documented K8s privilege escalation vectors referenced in CIS benchmarks and security audits

## New Checks

| Check | Severity | What it detects |
|-------|----------|----------------|
| `impersonate-without-resource-names` | High | Unrestricted impersonate verb enabling system:masters escalation |
| `escalate-verb-in-role` | High | Escalate verb bypassing RBAC escalation protection |
| `bind-verb-in-role` | High | Bind verb enabling arbitrary role binding |
| `approve-signers-without-resource-names` | High | Unrestricted certificate signer approval |
| `wildcard-resource-verb-combo` | Critical | Triple-wildcard rules functionally equivalent to cluster-admin |
| `aggregate-to-admin-escalation` | High | Aggregate-labeled roles inheriting escalation verbs |
| `deletecollection-verb-in-role` | High | Bulk resource destruction via deletecollection |
| `webhook-failure-policy-ignore` | High | Admission webhooks failing open |

## Test plan
- [ ] Each check has a test manifest with pass (dont-fire) and fail cases
- [ ] All checks use the existing cel-expression template — no new dependencies
- [ ] Run `go test ./...` to verify no regressions
- [ ] Run kube-linter against test manifests to verify check behavior

---
> **Note:** This PR was created with AI assistance using Claude Code.

### DIFF
--- a/e2etests/bats-tests.sh
+++ b/e2etests/bats-tests.sh
@@ -98,6 +98,42 @@ get_value_from() {
   [[ "${count}" == "1" ]]
 }
 
+@test "aggregate-to-admin-escalation" {
+  tmp="tests/checks/aggregate-to-admin-escalation.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include aggregate-to-admin-escalation --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+
+  count=$(get_value_from "${lines[0]}" '.Reports | length')
+  [[ "${count}" == "2" ]]
+}
+
+@test "approve-signers-without-resource-names" {
+  tmp="tests/checks/approve-signers-without-resource-names.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include approve-signers-without-resource-names --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+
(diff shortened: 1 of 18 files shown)


e2etests/bats-tests.sh
```diff
@@ -98,6 +98,42 @@ get_value_from() {
...
+@test "aggregate-to-admin-escalation" {
+  tmp="tests/checks/aggregate-to-admin-escalation.yml"
+  cmd="${KUBE_LINTER_BIN} lint --include aggregate-to-admin-escalation --do-not-auto-add-defaults --format json ${tmp}"
+  run ${cmd}
+
+  print_info "${status}" "${output}" "${cmd}" "${tmp}"
+  [ "$status" -eq 1 ]
+
+  count=$(get_value_from "${lines[0]}" '.Reports | length')
+  [[ "${count}" == "2" ]]
```
rhybrillou, line 110: (applies to all new tests here) Would it make sense to add an evaluation of the reports (extract message1 and message2, then compare them with expected values) ?

rhybrillou (changes requested):
Please review the CEL expression syntax (`has` does not support "complex" expressions such as map indexed lookup).

Please make sure the tests show the expected reported diagnostics, not just the expected count. It seems at the moment that some of the counts do not match the expectations, when comparing to the test input files (2 expected reports for 2 input objects in the `aggregate-to-admin-escalation` where one of the objects is called `dont-fire`).

PR #1232, opened 2026-07-29
test(memoryrequirements): add regression test for init+regular containers with limits set

### LINKED ISSUE #428: [BUG] unset-memory-requirements validation broken
**System info:**
 - OS: Windows 10 21H2

**Describe the bug**
When running `kube-linter` on a Helm Chart where initContainers and containers are defined, the validation rules fails even if the resources (limits and requests) in a Deployment YAMLs are defined. Removing the initContainers block, the validation works.

**To Reproduce**
Validate a Deployment YAML with both initContainers and containers brings to the error validation 

**Expected behavior**
The validation works and detects resources.

**Screenshots**
grafik

pujitha24:
Motivation:
Issue #428 reported that unset-memory-requirements produced a false
positive on a Deployment with both initContainers and containers when
resources were set on both. Reproducing this against current main with
a hand-built kube-linter binary and a Deployment YAML containing an
initContainer and a regular container, each with memory/cpu requests
and limits set, shows the check now correctly reports no violations.
The fix landed via earlier work (PodSpec.AllContainers(), used by
util.PerContainerCheck) that iterates init, regular, and ephemeral
containers while checking each container's own Resources field
independently. There was, however, no test locking in this behavior,
so the scenario could regress silently.

Approach:
- Add memoryrequirements/template_test.go using the repo's
  templates.TemplateTestSuite, covering:
  - an init container and a regular container both with memory
    requests/limits set -> no diagnostics (the exact #428 scenario).
  - an init container missing resources alongside a regular container
    that has them set -> the init container is still correctly
    flagged by name, proving the check isn't vacuously passing.
- Add AddInitContainerToDeployment to pkg/lintcontext/mocks/container.go,
  mirroring the existing AddContainerToDeployment helper, since no mock
  helper previously existed for populating a Deployment's init
  containers in tests.
- Extract the "memory-requirements" string literal into a templateKey
  const in template.go (matching the convention already used by
  sibling templates such as latesttag) so the new test can reference it
  via s.Init(templateKey).

Validation:
- go build ./...
- go test ./pkg/templates/memoryrequirements/... ./pkg/lintcontext/mocks/... (pass)
- go test ./... (all packages pass)
- golangci-lint run ./pkg/templates/memoryrequirements/... ./pkg/lintcontext/mocks/... (clean)

Impact: this is test-only coverage, not a functional fix. The
unset-memory-requirements/unset-cpu-requirements checks already behave
correctly on current main for the reported scenario; user-visible
behavior is unchanged. The benefit is a regression guard preventing
this specific false positive from silently reappearing.

Fixes #428

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/lintcontext/mocks/container.go
+++ b/pkg/lintcontext/mocks/container.go
@@ -15,3 +15,10 @@ func (l *MockLintContext) AddContainerToDeployment(t *testing.T, deploymentName 
 	// TODO: keep supporting other fields
 	deployment.Spec.Template.Spec.Containers = append(deployment.Spec.Template.Spec.Containers, container)
 }
+
+// AddInitContainerToDeployment adds a mock init container to the specified pod under context
+func (l *MockLintContext) AddInitContainerToDeployment(t *testing.T, deploymentName string, container v1.Container) {
+	deployment, ok := l.objects[deploymentName].(*appsV1.Deployment)
+	require.True(t, ok, "deployment with name %s not found", deploymentName)
+	deployment.Spec.Template.Spec.InitContainers = append(deployment.Spec.Template.Spec.InitContainers, container)
+}
--- a/pkg/templates/memoryrequirements/template.go
+++ b/pkg/templates/memoryrequirements/template.go
@@ -17,6 +17,8 @@ import (
 
 const (
 	bytesInMB = 1024 * 1024
+
+	templateKey = "memory-requirements"
 )
 
 func process(results *[]diagnostic.Diagnostic, containerName, requirementsType string, quantity *resource.Quantity, lowerBoundBytes int, upperBoundBytes *int) {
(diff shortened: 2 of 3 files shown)


pujitha24:
Good catch — the fixture set both `Requests` and `Limits`, but the test only validated with `RequirementsType: "limit"`, so request handling wasn't actually exercised. Updated `TestInitAndRegularContainersWithLimitsSet` to use `RequirementsType: "any"`, which checks both request and limit for the container per `template.go`. Confirmed it still passes with no diagnostics for the #428 regression scenario, and the full test suite (`go test ./...`) passes.

Re the failing `codecov/patch` and `codecov/project` checks: `codecov/project` appears to be comparing against a stale base commit (338 commits behind head) — the same large, bogus coverage drop shows up on unrelated open PRs (e.g. #1216) too, so that one looks like a pre-existing infra issue, not something from this diff. `codecov/patch` is flagging the new `AddInitContainerToDeployment` mock helper as uncovered, but that mirrors every other file in `pkg/lintcontext/mocks` (none have dedicated tests — they're exercised indirectly via other packages' tests, which the repo's `go test ./...` without `-coverpkg` doesn't attribute). I didn't want to unilaterally add a coverage-only test that breaks that convention, so leaving that as-is for a maintainer call.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

PR #1236, opened 2026-08-09
fix: report manifests that fail to decode instead of dropping them si…

### LINKED ISSUE #591: [BUG] Invalid value for `resources` results in misleading error message.
**System info:**
 - OS: [e.g. Linux? MaxOS? Windows?]
 - Linux
 - KubeLinter v0.6.2-0-g191de10fd5

**Describe the bug**
We are using [Terraform provider](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release) to deploy our charts. This allows us to provide some values at deployment time (usually infrastructure resources IDs etc.). Recently we've added alerts for pods resource consumption. To be DRY we've moved real values for `resources` to Terraform files from which they can be used for chart deployments and as alerts thresholds. To be able to template chart we still needed to have some values in `resources`  so we've used `"DEPLOY_TIME_VARIABLE"` like we do for other deployment time values. After this change kube-linter is failing with error message: `.../my-service/templates/service.yaml: (object: <no namespace>/my-service-api-svc /v1, Kind=Service) no pods found matching service labels (map[app:myservice]) (check: dangling-service, remediation: Confirm that your service's selector correctly matches the labels on one of your deployments.)` which seems to be not valid as the only change that was done was to values of the `resources`.

From: 

```yaml
# ./values.yaml

#...
app:
  replicaCount: 3
  resources:
    requests:
      cpu: 15m
      memory: 160Mi
    limits:
      cpu: 60m
      memory: 640Mi

#...
```

to:

```yaml
# ./values.yaml

#...
app:
  replicaCount: 3
  resources:
    requests:
      cpu: "DEPLOY_TIME_VARIABLE"
      memory: "DEPLOY_TIME_VARIABLE"
    limits:
      cpu: "DEPLOY_TIME_VARIABLE"
      memory: "DEPLOY_TIME_VARIABLE"
#...
```

**Sample YAML input**

```yaml
# ./Chart.yaml

apiVersion: v0.1
name: app-configuration-proxy
description: App Configuration Proxy Service
type: application
version: __RELEASE_TAG__
appVersion: __RELEASE_TAG__
```

```yaml
# ./values.yaml

image:
  registry: __ACR_NAME__
  repository: app-configuration-proxy
  tag: __RELEASE_TAG__

app:
  replicaCount: 3
  resources:
    requests:
      cpu: "DEPLOY_TIME_VARIABLE"
      memory: "DEPLOY_TIME_VARIABLE"
    limits:
      cpu: "DEPLOY_TIME_VARIABLE"
      memory: "DEPLOY_TIME_VARIABLE"

identity:
  type: 0
  resourceId: "DEPLOY_TIME_VARIABLE"
  clientId: "DEPLOY_TIME_VARIABLE"
  serviceAccountName: "DEPLOY_TIME_VARIABLE"

azureAd:
  tenantId: "DEPLOY_TIME_VARIABLE"
  domain: "DEPLOY_TIME_VARIABLE"
  clientId: "DEPLOY_TIME_VARIABLE"
  swaggerclientId: "DEPLOY_TIME_VARIABLE"
  groups:
    deploymentSlot: "DEPLOY_TIME_VARIABLE"
  allowToBeAuthorizedByAcl: "true"

keyvaultName: "DEPLOY_TIME_VARIABLE"

ingress:
  pathPrefix: "/app-configuration-proxy-service"
  defaultHost:
    dnsname: "DEPLOY_TIME_VARIABLE"
    defaultSecretName: "DEPLOY_TIME_VARIABLE"
  sitespecificHosts:
    enabled: false
    dnsZoneName: "DEPLOY_TIME_VARIABLE"
    sitecodes: ["DEPLOY_TIME_VARIABLE"]
    siteSpecificSecretName: "DEPLOY_TIME_VARIABLE"
  wafPolicyId: "DEPLOY_TIME_VARIABLE"

appInfo:
  applicationName: "DEPLOY_TIME_VARIABLE"
  deploymentStampName: "DEPLOY_TIME_VARIABLE"
  slotName: "DEPLOY_TIME_VARIABLE"
  serviceName: "DEPLOY_TIME_VARIABLE"
  environment: "DEPLOY_TIME_VARIABLE"
  component: "DEPLOY_TIME_VARIABLE"

azureAppConfigEndpoint: "DEPLOY_TIME_VARIABLE"

exposeSwagger: "EXPOSE_SWAGGER"
swaggerDescription: __SWAGGER_DESCRIPTION__
swaggerContactEmail: __SWAGGER_CONTACT_EMAIL__

applicationInsights:
  logLevel:
    default: __ApplicationInsights_LogLevel_Default__

logging:
  logLevel:
    default: __Logging_LogLevel_Default__

secretstore:
  deploymentStamp:
    keyvaultName: kv-__APP_NAME__-app-__ENVIRONMENT__
    secrets:
      - kvName: "auto-acpx-sp-client-secret"
        envName: "AzureAd__Secret"

defaultDomain:
  dtlsKeyvaultCertName: my-cert-name

sitespecificDomain:
  enabled: false
  siteTlsKeyvaultCertName: my-cert-name

```

```tpl
# ./templates/_helpers.tpl

{{- define "app-configuration-proxy-service.name" -}}
app-configuration-proxy
{{- end }}

{{- define "app-configuration-proxy-service.fullname" -}}
{{ include "app-configuration-proxy-service.name" . }}
{{- end }}

{{- define "app-configuration-proxy-service.chart" -}}
{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
{{- end }}

{{- define "app-configuration-proxy-service.labels" -}}
helm.sh/chart: {{ include "app-configuration-proxy-service.chart" . }}
{{ include "app-configuration-proxy-service.selectorLabels" . }}
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
{{- end }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end }}

{{- define "app-configuration-proxy-service.selectorLabels" -}}
app.kubernetes.io/name: {{ include "app-configuration-proxy-service.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end }}

{{- define "app-configuration-proxy-service.serviceAccountName" -}}
{{- if .Values.serviceAccount.create }}
{{- default (include "app-configuration-proxy-service.fullname" .) .Values.serviceAccount.name }}
{{- else }}
{{- default "default" .Values.serviceAccount.name }}
{{- end }}
{{- end }}

```

```yaml
# ./templates/deployment-stamp-kv-secrets.yaml

apiVersion: secrets-store.csi.x-k8s.io/v1
kind: SecretProviderClass
metadata:
  name: "{{ include "app-configuration-proxy-service.name" . }}-stamp-secretproviderclass"
spec:
  provider: azure
  secretObjects:
  - data:
      {{- range .Values.secretstore.deploymentStamp.secrets }}
      - objectName: {{ .kvName }}
        key: {{ .kvName }}
      {{- end }}
    secretName: "{{ include "app-configuration-proxy-service.name" . }}-stamp-secrets"
    type: Opaque

  parameters:
    usePodIdentity: "false"
    useVMManagedIdentity: "false"
    clientID: "{{ .Values.identity.clientId }}"
    keyvaultName: "{{ .Values.secretstore.deploymentStamp.keyvaultName }}"
    cloudName: ""
    objects:  |
      array:
        {{- range .Values.secretstore.deploymentStamp.secrets }}
        - |
          objectName: {{ .kvName }}
          objectType: secret
          objectVersion: ""
        {{- end }}
    tenantId: "{{ .Values.azureAd.tenantId }}"

```

```yaml
# ./templates/deployment.yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "app-configuration-proxy-service.fullname" . }}-deployment
  labels:
    app: appconfigurationproxyservice
    {{- include "app-configuration-proxy-service.labels" . | nindent 4 }}
spec:
  replicas: {{ .Values.app.replicaCount }}
  revisionHistoryLimit: 5
  selector:
    matchLabels:
      app: appconfigurationproxyservice
  template:
    metadata:
      labels:
        app: appconfigurationproxyservice
        azure.workload.identity/use: "true"
        {{- include "app-configuration-proxy-service.labels" . | nindent 8 }}
    spec:
      serviceAccountName: {{ .Values.identity.serviceAccountName }}
      securityContext:
        runAsUser: 1000
        runAsGroup: 2000
        fsGroup: 2000
      affinity:
        podAntiAffinity:
          - weight: 100
            podAffinityTerm:
              labelSelector:
                matchExpressions:
                - key: app
                  operator: In
                  values:
                  - appconfigurationproxyservice
      containers:
      - name: app-configuration-proxy
        image: {{ .Values.image.registry }}.azurecr.io/{{ .Values.image.repository }}:{{ .Values.image.tag }}
        securityContext:
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
        resources:
          requests:
            cpu: {{ .Values.app.resources.requests.cpu }}
            memory: {{ .Values.app.resources.requests.memory }}
          limits:
            cpu: {{ .Values.app.resources.limits.cpu }}
            memory: {{ .Values.app.resources.limits.memory }}
        ports:
        - containerPort: 8080
        envFrom:
        - secretRef:
            name: {{ include "app-configuration-proxy-service.fullname" . }}-azuread
        env:
          - name: PathPrefix
            value: {{ .Values.ingress.pathPrefix }}
          - name: ExposeSwagger
            value: {{ .Values.exposeSwagger | quote }}
          - name: SwaggerDescription
            value: {{ .Values.swaggerDescription | quote }}
          - name: SwaggerContactEmail
            value: {{ .Values.swaggerContactEmail | quote }}
          - name: APPINSIGHTS_INSTRUMENTATIONKEY
            value: {{ .Values.app_insights_instrumentation_key }}
          - name: AppInfo__ApplicationName
            value: {{ .Values.appInfo.applicationName }}
          - name: AppInfo__DeploymentStampName
            value: {{ .Values.appInfo.deploymentStampName }}
          - name: AppInfo__SlotName
            value: {{ .Values.appInfo.slotName }}
          - name: AppInfo__ServiceName
            value: {{ .Values.appInfo.serviceName }}
          - name: AppInfo__Environment
            value: {{ .Values.appInfo.environment }}
          - name: AppInfo__Component
            value: {{ .Values.appInfo.component }}
          - name: AzureAppConfigEndpoint
            value: {{ .Values.azureAppConfigEndpoint }}
          {{- range .Values.secretstore.deploymentStamp.secrets}}
          - name: {{ .envName}}
            valueFrom:
              secretKeyRef:
                name: "{{ include "app-configuration-proxy-service.name" . }}-stamp-secrets"
                key: {{ .kvName}}
          {{- end }}
        volumeMounts:
          - name: secrets-store-stamp
            mountPath: /mnt/secrets-stamp
            readOnly: true
          - name: tmp
            mountPath: /tmp
        readinessProbe:
          httpGet:
            path: /health/ready
            port: 8080
          periodSeconds: 3
          timeoutSeconds: 1
        livenessProbe:
          httpGet:
            path: /health/live
            port: 8080
          periodSeconds: 3
          timeoutSeconds: 1
      volumes:
      - name: secrets-store-stamp
        csi:
          driver: secrets-store.csi.k8s.io
          readOnly: true
          volumeAttributes:
            secretProviderClass: "{{ include "app-configuration-proxy-service.name" . }}-stamp-secretproviderclass"
      - name: tmp
        emptyDir: {}
```

```yaml
# ./templates/ingress.yaml

apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: {{ include "app-configuration-proxy-service.fullname" . }}-api-ingress
  annotations:
    kubernetes.io/ingress.class: azure/application-gateway
    appgw.ingress.kubernetes.io/backend-path-prefix: "/"
    appgw.ingress.kubernetes.io/ssl-redirect: "true"
    appgw.ingress.kubernetes.io/waf-policy-for-path: "{{ .Values.ingress.wafPolicyId }}"
spec:
  rules:
  - host: "{{ .Values.ingress.defaultHost.dnsname }}"
    http:
      paths:
      - path: "{{ .Values.ingress.pathPrefix }}/*"
        pathType: ImplementationSpecific
        backend:
          service:
            name: {{ include "app-configuration-proxy-service.fullname" . }}-api-svc
            port:
              number: 80
  {{- if .Values.ingress.sitespecificHosts.enabled }}
  {{- range .Values.ingress.sitespecificHosts.sitecodes }}
  - host: "{{ . }}.{{ $.Values.ingress.sitespecificHosts.dnsZoneName }}"
    http:
      paths:
        - path: "{{ $.Values.ingress.pathPrefix }}/*"
          pathType: ImplementationSpecific
          backend:
            service:
              name: {{ include "app-configuration-proxy-service.fullname" . }}-api-svc
              port:
                number: 80
  {{- end }}
  {{- end }}
  tls:
  - hosts:
    - "{{ .Values.ingress.defaultHost.dnsname }}"
    secretName: {{ .Values.ingress.defaultHost.defaultSecretName }}
  {{- if .Values.ingress.sitespecificHosts.enabled }}
  - hosts:
    {{- range .Values.ingress.sitespecificHosts.sitecodes }}
    - "{{ . }}.{{ $.Values.ingress.sitespecificHosts.dnsZoneName }}"
    {{- end }}
    secretName: {{ .Values.ingress.sitespecificHosts.siteSpecificSecretName }}
  {{- end }}

```

```yaml
# ./templates/secrets.yaml

apiVersion: v1
kind: Secret
metadata:
  name: {{ include "app-configuration-proxy-service.fullname" . }}-azuread
type: Opaque
data:
  AzureAd__Instance: {{ printf "https://login.microsoftonline.com" | b64enc | quote }}
  AzureAd__TenantId: {{ .Values.azureAd.tenantId | b64enc | quote }}
  AzureAd__Domain: {{ .Values.azureAd.domain | b64enc | quote }}
  AzureAd__ClientId: {{ .Values.azureAd.clientId | b64enc | quote }}
  AzureAd__Groups__DeploymentSlot: {{ .Values.azureAd.groups.deploymentSlot | b64enc | quote }}
  AzureAd__AllowWebApiToBeAuthorizedByACL: {{ .Values.azureAd.allowToBeAuthorizedByAcl | b64enc | quote }}
  SwaggerUI__ClientId: {{ .Values.azureAd.swaggerclientId | b64enc | quote }}

```

```yaml
# ./templates/service.yaml

apiVersion: v1
kind: Service
metadata:
  name: {{ include "app-configuration-proxy-service.fullname" . }}-api-svc
spec:
  selector:
    app: appconfigurationproxyservice
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8080

```

```yaml
# ./templates/serviceaccounts.yaml

apiVersion: v1
kind: ServiceAccount
metadata:
  annotations:
    azure.workload.identity/client-id: {{ .Values.identity.clientId }}
  labels:
    azure.workload.identity/use: "true"
  name: {{ .Values.identity.serviceAccountName }}
automountServiceAccountToken: true

```

**Expected behavior**

* Error messages which would clearly indicate root-cause of the issue.
* It should be possible to configure linter to accept any value for `resources` including placeholders and only warn user if it cannot check if the value seems to be reasonable but it should not count it as error.

### LINKED ISSUE #669: [Bug] no pods found matching service labels
image

After scanning a file, linter reports that it cannot find pods by labels, although the labels in the deploy and service are correct.

### web.yaml
```
apiVersion: v1
kind: Service
metadata:
  name: admin
spec:
  selector:
    app.name: ivea-django
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8000
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: admin
spec:
  replicas: 1
  selector:
    matchLabels:
      app.name: ivea-django
  template:
    metadata:
      labels:
        app.name: ivea-django
      annotations:
        lastDeployedDate: "{{ now | unixEpoch }}"
    spec:
      imagePullSecrets:
      - name: gitlab-registry-credentials
      containers:
      - name: main
        imagePullPolicy: Always
        image: {{ .Values.images.admin }}
---
apiVersion: batch/v1
kind: Job
metadata:
  name: admin-migrations-job-{{ .Release.Revision }}
  annotations:
    helm.sh/hook: pre-install,pre-upgrade
spec:
  backoffLimit: 0
  ttlSecondsAfterFinished: 3600
  template:
    spec:
      restartPolicy: Never
      imagePullSecrets:
        - name: gitlab-registry-credentials
      containers:
        - name: main
          image: {{ .Values.images.admin }}
          imagePullPolicy: Always
          command: ["python3", "manage.py", "migrate"]
          envFrom:
            - secretRef:
                name: admin
```

Eljees:
…lently

parseObjects() falls back to an unstructured object whenever the typed decoder fails. The fallback exists for custom resources whose Go type KubeLinter does not know, but it also swallowed a manifest of a registered kind that simply is invalid: the workload became an unstructured object, extract.PodTemplateSpec() no longer saw it, and every pod-spec based check silently stopped applying to it. The user was left with an unrelated diagnostic about a different object.

Restrict the fallback to kinds that are not registered in the scheme (runtime.IsNotRegisteredError), so a broken Deployment surfaces as an invalid object while custom resources keep parsing as before.

The invalid objects were then still hidden behind --verbose in the lint command; that gate arrived with a mechanical code move and is dropped, so the load failure and its cause are printed by default.

Fixes #591
Fixes #669

### DIFF
--- a/pkg/lintcontext/parse_yaml.go
+++ b/pkg/lintcontext/parse_yaml.go
@@ -68,6 +68,13 @@ func parseObjects(data []byte, d runtime.Decoder) ([]k8sutil.Object, error) {
 		if strings.Contains(err.Error(), "json: cannot unmarshal") {
 			return nil, fmt.Errorf("failed to decode: %w", err)
 		}
+		// The fallback below exists for custom resources, whose Go types we do not know.
+		// If the kind is registered, the manifest is simply invalid: decoding it as
+		// unstructured would swallow the real error and, worse, hide the object from every
+		// check that relies on its typed representation (see extract.PodTemplateSpec).
+		if !runtime.IsNotRegisteredError(err) {
+			return nil, fmt.Errorf("failed to decode: %w", err)
+		}
 		// fallback to unstructured as schema validation will be performed by kubeconform check
 		dec := runtimeYaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
 		var unstructuredErr error
--- a/pkg/lintcontext/parse_yaml.go
+++ b/pkg/lintcontext/parse_yaml.go
@@ -184,13 +184,28 @@ func (l *lintContextImpl) renderTgzHelmChart(tgzFile string) (map[string]string,
 	return l.renderChart(tgzFile, chrt)
 }
 
+// isBlankDocument reports whether a YAML document carries no content at all: every
+// line is blank, a comment, or a bare document marker. Files that separate sections
+// with "--- # some comment" produce such documents, and there is nothing in them to
+// decode - reporting them as unreadable objects would be a false positive.
+func isBlankDocument(doc []byte) bool {
+	for _, line := range bytes.Split(doc, []byte("\n")) {
+		line = bytes.TrimSpace(line)
+		if len(line) == 0 || line[0] == '#' || bytes.Equal(line, []byte("---")) || bytes.Equal(line, []byte("...")) {
+			continue
+		}
+		return false
+	}
+	return true
+}
+
 func (l *lintContextImpl) loadObjectFromYAMLReader(filePath string, r *yaml.YAMLReader) error {
 	doc, err := r.Read()
 	if err != nil {
 		return err
 	}
 	doc = bytes.TrimSpace(doc)
-	if len(doc) == 0 {
+	if len(doc) == 0 || isBlankDocument(doc) {
 		return nil
 	}
 
--- a/e2etests/invalid_object_test.go
+++ b/e2etests/invalid_object_test.go
@@ -1,0 +1,98 @@
+//go:build e2e
+
+package e2etests
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Regression test for #591 / #669.
+//
+// A manifest of a known kind that fails to decode used to be silently downgraded to an
+// unstructured object. It then became invisible to every pod-spec based check, and the only
+// thing the user saw was an unrelated dangling-service diagnostic about a *different* object.
+// KubeLinter must say which file it could not read, without needing --verbose.
+func TestKubeLinterReportsUndecodableObjectsWithoutVerbose(t *testing.T) {
+	kubeLinterBin := os.Getenv(kubeLinterBinEnv)
+	require.NotEmpty(t, kubeLinterBin, "Please set %s", kubeLinterBinEnv)
+
+	cases := map[string]struct {
+		manifest    string
+		wantInError string
+	}{
+		// #591: an invalid resource quantity.
+		"invalid quantity": {
+			manifest: `apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+spec:
+  selector:
+    matchLabels:
+      name: my-label-value
+  template:
+    metadata:
+      labels:
+        name: my-label-value
+    spec:
+      containers:
+        - name: my-container
+          image: nginx:latest
+          resources:
+            limits:
+              cpu: 10wrong_unit
+`,
+			wantInError: "quantities must match",
+		},
+		// #669: a number where the schema wants a string.
+		"number instead of string": {
+			manifest: `apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+spec:
+  selector:
+    matchLabels:
+      name: my-label-value
+  template:
+    metadata:
+      labels:
+        name: my-label-value
+    spec:
+      containers:
+        - name: my-container
+          image: nginx:latest
+          livenessProbe:
+            exec:
+              command:
+                - /usr/bin/echo
+                - 1234
+`,
+			wantInError: "cannot unmarshal number",
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
+			dir := t.TempDir()
+			path := filepath.Join(dir, "manifest.yaml")
+			require.NoError(t, os.WriteFile(path, []byte(c.manifest), 0644))
+
+			// Deliberately no --verbose: the user must be told anyway.
+			out, _ := exec.Command(kubeLinterBin, "lint", path).CombinedOutput()
+			outAsStr := string(out)
+
+			assert.True(t, strings.Contains(outAsStr, "failed to load object"),
+				"expected a load failure to be reported, got: %s", outAsStr)
+			assert.True(t, strings.Contains(outAsStr, c.wantInError),
+				"expected the report to name the real cause %q, got: %s", c.wantInError, outAsStr)
+		})
+	}
+}
--- a/pkg/command/lint/command.go
+++ b/pkg/command/lint/command.go
@@ -105,10 +105,11 @@ func Command() *cobra.Command {
 				return err
 			}
 			invalidObjectsResult := generateReportFromInvalidObjects(lintCtxs)
-			if verbose {
-				for _, invalidObj := range invalidObjectsResult {
-					_, _ = fmt.Fprintf(os.Stderr, "Warning: failed to load object from %s: %v\n", invalidObj.Object.Metadata.FilePath, invalidObj.Diagnostic.Message)
-				}
+			// Objects we failed to load are silently missing from every check that would
+			// have applied to them, which surfaces as an unrelated diagnostic somewhere
+			// else (see #591). Always tell the user what we could not read.
+			for _, invalidObj := range invalidObjectsResult {
+				_, _ = fmt.Fprintf(os.Stderr, "Warning: failed to load object from %s: %v\n", invalidObj.Object.Metadata.FilePath, invalidObj.Diagnostic.Message)
 			}
 
 			var atLeastOneObjectFound = errorOnInvalidResource && len(invalidObjectsResult) > 0
--- a/pkg/lintcontext/parse_yaml_invalid_object_test.go
+++ b/pkg/lintcontext/parse_yaml_invalid_object_test.go
@@ -1,6 +1,7 @@
 package lintcontext
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -89,4 +90,39 @@ spec:
 	require.NoError(t, err)
 	require.Len(t, objects, 1)
 	assert.IsType(t, &appsV1.Deployment{}, objects[0])
+}
+
+// Reporting unreadable manifests must not turn section separators into findings.
+// A document that holds nothing but comments carries no manifest at all, so it is
+// neither a valid object nor an invalid one.
+func TestLoadObjectsFromReaderSkipsCommentOnlyDocuments(t *testing.T) {
+	const doc = `---
+# Secret test cases
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+spec:
+  selector:
+    matchLabels:
+      name: my-label-value
+  template:
+    metadata:
+      labels:
+        name: my-label-value
+    spec:
+      containers:
+        - name: my-container
+          image: nginx:latest
+---
+# ConfigMap test cases
+---
+`
+
+	ctx := newCtx(Options{})
+	require.NoError(t, ctx.loadObjectsFromReader("test.yaml", strings.NewReader(doc)))
+	assert.Empty(t, ctx.InvalidObjects(), "comment-only documents must not be reported as unreadable")
+	require.Len(t, ctx.Objects(), 1)
+	assert.Equal(t, "my-deployment", ctx.Objects()[0].K8sObject.GetName())
 }
--- a/pkg/lintcontext/parse_yaml_invalid_object_test.go
+++ b/pkg/lintcontext/parse_yaml_invalid_object_test.go
@@ -1,0 +1,92 @@
+package lintcontext
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	appsV1 "k8s.io/api/apps/v1"
+)
+
+// Regression test for #591: a manifest of a *known* kind that fails to decode must be
+// reported as an invalid object instead of being silently downgraded to Unstructured.
+// The downgrade makes the workload invisible to every pod-spec based check, so the user
+// gets an unrelated diagnostic (dangling-service) and never learns about the real problem.
+func TestParseObjectsKnownKindWithInvalidValueIsReported(t *testing.T) {
+	const brokenDeployment = `apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: my-deployment
+spec:
+  selector:
+    matchLabels:
+      name: my-label-value
+  template:
+    metadata:
+      labels:
+        name: my-label-value
+    spec:
+      containers:
+        - name: my-container
+          image: nginx:latest
+          resources:
+            limits:
+              cpu: 10wrong_unit
+`
+
+	objects, err := parseObjects([]byte(brokenDeployment), nil)
+	assert.Error(t, err, "a Deployment with an invalid quantity must not parse silently")
+	assert.Empty(t, objects)
+	if err != nil {
+		assert.Contains(t, err.Error(), "quantities must match", "the error must name the real cause")
+	}
+}
+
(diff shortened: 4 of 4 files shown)


Eljees:
The first CI run was red on `e2e-bats`. Printing load failures by default made the repository's own `tests/checks/env-var-value-from.yml` emit one warning per comment-only document, so `${lines[0]}` in the bats helper was a warning line instead of the JSON payload and `jq` bailed out.

That turned out to be a real false positive rather than a test problem: a document that holds nothing but comments is not a manifest, yet it was already being recorded as an invalid object — and already failed `--error-on-invalid-resource` — it was simply hidden behind `--verbose` until this PR.

Fixed in 30ed792: a document whose every line is blank, a comment, or a bare document marker is now skipped before decoding, the same way an empty one is. Added a regression test for it; no existing test or fixture was adjusted to accommodate the change.

Eljees:
The red check here is `codecov/project`, and it is not about this branch.

Codecov's chosen base is `dbd7529` - "chore(deps): bump helm.sh/helm/v3 from 3.15.1 to 3.15.2 (#796)", 13 June 2024, 342 commits behind this PR's head. So the comparison counts two years of intervening work as this patch: it reports 1,959 patch lines against the 254 this branch actually adds, and a base of 197 files against a head of 239.

It fails the same way on every open PR here that has a codecov context - #1189, #1197, #1206, #1216 - and passes on none of them. `codecov/patch` is green on this one.

The pattern fits coverage uploads having stopped landing on `main` at some point, so codecov keeps falling back to the last commit it holds a report for. Happy to look at the upload step separately if that would be useful; it is outside this fix.

pkg/lintcontext/parse_yaml.go
```diff
@@ -177,13 +184,28 @@ func (l *lintContextImpl) renderTgzHelmChart(tgzFile string) (map[string]string,
...
 }
 
+// isBlankDocument reports whether a YAML document carries no content at all: every
+// line is blank, a comment, or a bare document marker. Files that separate sections
+// with "--- # some comment" produce such documents, and there is nothing in them to
+// decode - reporting them as unreadable objects would be a false positive.
+func isBlankDocument(doc []byte) bool {
+	for _, line := range bytes.Split(doc, []byte("\n")) {
+		line = bytes.TrimSpace(line)
+		if len(line) == 0 || line[0] == '#' || bytes.Equal(line, []byte("---")) || bytes.Equal(line, []byte("...")) {
```
rhybrillou, line 194: Would it make sense to consider the cases where the start and end document markers are combined with comment ?
```suggestion
		if bytes.HasPrefix(line, []byte("---") {
			line = bytes.TrimPrefix(line, []byte("---")
			line = bytes.TrimSpace(line)
		} else if bytes.HasPrefix(line, []byte("...") {
			line = bytes.TrimPrefix(line, []byte("...")
			line = bytes.TrimSpace(line)
		}
		if len(line) == 0 || line[0] == '#' {
```
That would address one of the coderabbit comments.

rhybrillou (commented):
Please adjust the parsing of multi-document files.
The rest of the change looks good.

Eljees:
Done in `bf47d3a`.

`isBlankDocument` now strips a leading `---` or `...` from the line and judges what is left: nothing or a comment leaves the document empty, anything else stays content and goes to the decoder.

One thing worth recording about which half of this the parser actually sees. The k8s YAML reader handles the start marker itself — `--- # section two` is consumed as a separator, and it rejects anything but a comment after the marker — so that line never reaches `isBlankDocument`. The end marker is different: `... # end of section` is not a separator, so it arrives as the entire body of a document. That is the case that produced a false invalid object here, and the one the change removes; the start marker is handled the same way for symmetry, and because `isBlankDocument` is reachable from other callers.

Two tests, both of which fail on the previous implementation and pass on this one:

* `TestLoadObjectsFromReaderSkipsDocumentMarkersWithComments` goes through `loadObjectsFromReader` with commented markers around a real Deployment. Without the change the trailing `... # end of section` document is reported as unreadable.
* `TestIsBlankDocumentKeepsContentAfterMarker` is a table over the predicate itself and pins both directions, including the cases that must stay content: `--- !SomeTag` and `---foo: bar`.

I confirmed the failure direction by reverting only the change and keeping both tests: `TestLoadObjectsFromReaderSkipsDocumentMarkersWithComments` and two rows of the table fail, and every other test in the package still passes. `gofmt`, `go vet` and `go test ./pkg/lintcontext/` are clean on the new head with Go 1.26.4.

The remaining red check is `codecov/project`, which compares against a base 342 commits behind this head and fails the same way on every open PR here — the detail is in my comment above.

rhybrillou (approved):

PR #1243, opened 2026-08-25
feat: add hpa-maximum-replicas check

### LINKED ISSUE #608: [FEATURE_REQUEST] Support HPA maxReplicas
Add check for HPA maxReplicas which functions in the same way as minReplicas

**Additional context**
We are migrating from datree and this was a check they have: https://hub.datree.io/built-in-rules/ensure-hpa-maximum-replicas

ChrisJr404:
## Summary

Closes #608.

This adds an `hpa-maximum-replicas` check that flags HorizontalPodAutoscaler objects whose `maxReplicas` is allowed to climb past a configured ceiling. It works the same way as `hpa-minimum-replicas`, just from the other side, so the two can be paired to keep a HorizontalPodAutoscaler inside a replica range.

I kept it as a separate template instead of folding a max param into the existing one, so the minimum check keeps behaving exactly as before (the backward compatibility point raised in the issue thread). The new check is not enabled by default, since a reasonable ceiling depends on the cluster.

It covers the standard autoscaling versions plus KEDA ScaledObjects (via the existing `GetHPAMaxReplicas` helper), and it comes with unit tests, an e2e test and generated docs, matching how the minimum check is wired up.


pkg/templates/hpamaxreplicas/template.go
```diff
@@ -0,0 +1,48 @@
...
+	templateKey = "hpa-maximum-replicas"
+)
+
+func init() {
+	templates.Register(check.Template{
+		HumanName:   "HorizontalPodAutoscaler Maximum replicas",
+		Key:         templateKey,
+		Description: "Flag applications running more than the specified number of replicas",
+		SupportedObjectKinds: config.ObjectKindsDesc{
+			ObjectKinds: []string{objectkinds.HorizontalPodAutoscaler},
```
rhybrillou, line 27:
```suggestion
			ObjectKinds: []string{
				objectkinds.HorizontalPodAutoscaler,
				objectkinds.ScaledObject,
			},
```
This should partially fix the coderabbit comment below.

pkg/builtinchecks/yamls/hpa-maximum-replicas.yaml
```diff
@@ -0,0 +1,10 @@
+name: "hpa-maximum-replicas"
+description: "Indicates when a HorizontalPodAutoscaler specifies more than the allowed maxReplicas"
+remediation: >-
+  Lower the maxReplicas in the HorizontalPodAutoscaler, or raise the check's maxReplicas parameter, so that autoscaling stays within the capacity you intend to allow.
+scope:
+  objectKinds:
+    - HorizontalPodAutoscaler
```
rhybrillou, line 7:
```suggestion
    - HorizontalPodAutoscaler
    - ScaledObject
```

pkg/templates/hpamaxreplicas/template_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"golang.stackrox.io/kube-linter/pkg/lintcontext/mocks"
+	"golang.stackrox.io/kube-linter/pkg/templates"
+	"golang.stackrox.io/kube-linter/pkg/templates/hpamaxreplicas/internal/params"
+	autoscalingV1 "k8s.io/api/autoscaling/v1"
+	autoscalingV2 "k8s.io/api/autoscaling/v2"
+	autoscalingV2Beta1 "k8s.io/api/autoscaling/v2beta1"
+	autoscalingV2Beta2 "k8s.io/api/autoscaling/v2beta2"
+)
+
+var autoscalingVersions = [4]string{"v2beta1", "v2beta2", "v2", "v1"}
```
rhybrillou, line 18: Could you add a test for the KEDA ScaleObject type ?

pkg/templates/hpamaxreplicas/template.go
```diff
@@ -0,0 +1,48 @@
...
+		},
+		Parameters:             params.ParamDescs,
+		ParseAndValidateParams: params.ParseAndValidate,
+		Instantiate: params.WrapInstantiateFunc(func(p params.Params) (check.Func, error) {
+			return func(_ lintcontext.LintContext, object lintcontext.Object) []diagnostic.Diagnostic {
+				replicas, found := extract.HPAMaxReplicas(object.K8sObject)
+				if !found {
+					return nil
+				}
+				if int(replicas) <= p.MaxReplicas {
```
rhybrillou, line 37: Would it make sense to deactivate the check if p.MaxReplicas is <= 0 ?
Would it make sense to ensure that replicas >= 0 ?

Issue #1242, opened 2026-08-23
codecov/project fails on every PR: coverage uploads from main have been rejected since June 2024


Eljees:
`codecov/project` is red on every open pull request that has a codecov context - #1189, #1197, #1206, #1216, #1236 - and green on none of them. The cause is not in those branches.

## Codecov has had no report for `main` since June 2024

`api.codecov.io/api/v2/github/stackrox/repos/kube-linter/commits/?branch=main` returns 19 commits, the newest being `dbd7529` - "chore(deps): bump helm.sh/helm/v3 from 3.15.1 to 3.15.2 (#796)", 13 June 2024.

Every PR is therefore compared against a base 340+ commits behind, and two years of intervening work counts as the patch. On #1236 codecov reports 1,959 patch lines against the 254 that branch actually adds, and a base of 197 files against a head of 239.

## The uploads are being rejected, silently

From the last `push`-triggered run on `main` ([run 28674157419](https://github.com/stackrox/kube-linter/actions/runs/28674157419), 3 July 2026, `build-and-test`):

```
info  - 2026-07-03 17:18:30,985 -- Found 1 coverage files to report
info  - 2026-07-03 17:18:30,985 -- > /home/runner/work/kube-linter/kube-linter/coverage.out
info  - 2026-07-03 17:18:31,615 -- Upload queued for processing complete
error - 2026-07-03 17:18:31,615 -- Upload queued for processing failed: {"message":"Repository not found"}
```

The step still reports success, because `fail_ci_if_error` defaults to false. That is why this has gone unnoticed for so long.

"Repository not found" is what codecov returns for an upload token that does not match the repository, and the rest of the evidence fits: the branch list on codecov contains only fork branches - `pujitha24:...`, `ugiordan:...`, `mwozniakk:...`, `bdsoha:...`, `arpitjain099:...` - and no in-repo branch at all. Pull requests from forks have no access to `secrets.CODECOV_TOKEN`, so they upload tokenless and succeed; the runs that do have the secret send a token codecov rejects.

## A second gap, independent of the token

`.github/workflows/build.yaml` has not run on a `push` to `main` since 3 July, yet `main` has had 12 commits since. All of them are dependabot merges, and `.github/workflows/auto-merge.yaml` performs those with

```yaml
run: gh pr merge --auto --squash "$PR_URL" && gh pr review --approve "$PR_URL"
env:
  GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```

[...]

PR #1262, opened 2026-09-21
fix(dangling-servicemonitor): an empty namespaceSelector means the monitor's own namespace

### LINKED ISSUE #1256: [BUG] dangling-servicemonitor ignores the monitor namespace by default
**System info:**

- OS: macOS 26.5.1, arm64
- kube-linter: v0.8.3
- Also confirmed in current `main` source at `ea3abb8abfe0fc4d7c2b20c9a30d95db9f0b2e2a`

**Describe the bug**

The `dangling-servicemonitor` check accepts a Service from a different namespace when the ServiceMonitor omits `spec.namespaceSelector`.

Prometheus Operator defines an empty `NamespaceSelector` as selecting objects from the current namespace. In this example, the ServiceMonitor is in `monitoring`, so the Service in `application` should not satisfy it.

**To Reproduce**

Save the YAML below as `repro.yaml`, then run:

```console
kube-linter lint repro.yaml --do-not-auto-add-defaults --include dangling-servicemonitor
```

The v0.8.3 binary exits 0 with:

```text
KubeLinter 0.8.3

No lint errors found!
```

**Sample YAML input**

```yaml
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: application-metrics
  namespace: monitoring
spec:
  selector:
    matchLabels:
      app: application
  endpoints:
    - port: metrics
---
apiVersion: v1
kind: Service
metadata:
  name: application
  namespace: application
  labels:
    app: application
spec:
  selector:
    app: application
  ports:
    - name: metrics
      port: 9090
      targetPort: metrics
```

**Expected behavior**

The check should report `application-metrics` as dangling because its empty namespace selector restricts discovery to `monitoring`, where no matching Service exists.

Two controls behave as expected:

- Moving the Service to `monitoring` produces no finding.
- Adding `namespaceSelector.matchNames: [application]` to the ServiceMonitor produces no finding.

**Additional context**

The current implementation marks the namespace selector as unset, then [returns success for any label-matching Service](https://github.com/stackrox/kube-linter/blob/ea3abb8abfe0fc4d7c2b20c9a30d95db9f0b2e2a/pkg/templates/danglingservicemonitor/template.go#L35-L68), regardless of that Service's namespace. The Prometheus Operator API states that empty `matchNames` with `any: false` [selects from the current namespace](https://github.com/prometheus-operator/prometheus-operator/blob/ff5cad994c539b3568b314f3c5086275e93800f4/pkg/apis/monitoring/v1/types.go#L792-L805).

A focused fix would preserve the existing `any` and `matchNames` behavior while treating an empty namespace selector as the ServiceMonitor's namespace. A regression test with the explicit monitor and Service namespaces above should cover the false negative.

As an independent relationship cross-check, public [`iac-guard-v==0.1.0b1`](https://pypi.org/project/iac-guard-v/0.1.0b1/) reports `VIOLATED` with `SERVICEMONITOR_TARGET_UNRESOLVED` for this case, and `SATISFIED` for both controls.

Eljees:
Fixes #1256.

### The defect

Prometheus Operator reads an empty `NamespaceSelector` as "the namespace this ServiceMonitor is in". `dangling-servicemonitor` had one branch that never looked at namespaces at all:

```go
if labelSelector.Matches(labels.Set(services.Labels)) && labelSelectorSet && !nsSelectorSet {
    // Found!
    return nil
}
```

A Service with matching labels in any namespace satisfied a monitor that set no namespace selector, so the dangling monitor in the report went unreported.

### The change

The namespace is compared on that branch. A manifest that leaves the namespace out on either side carries nothing to compare - a Helm template rendered without one, or a file meant to be applied with `-n` - and keeps matching as it did. That shape is not hypothetical: the `dont-fire` pair in `tests/checks/dangling-servicemonitor.yml` is exactly it, a Service in `dontfire` with a ServiceMonitor that has no namespace, and it must keep passing.

### Red, then green

The new unit case fails on untouched `main`:

```
--- FAIL: TestDanglingServiceMonitor/TestEmptyNamespaceSelectorDoesNotReachAnotherNamespace
    expected diagnostics and actual diagnostics do not match
    extra elements in list A:
     "no services found matching the service monitor's label selector (app=service1-test) and namespace selector ([])"
```

With the change:

```
$ go test ./pkg/templates/danglingservicemonitor/...
--- PASS: TestDanglingServiceMonitor/TestDanglingNamespaceSelector
--- PASS: TestDanglingServiceMonitor/TestEmptyNamespaceSelectorDoesNotReachAnotherNamespace
--- PASS: TestDanglingServiceMonitor/TestEmptyNamespaceSelectorMatchesItsOwnNamespace
--- PASS: TestDanglingServiceMonitor/TestInvalidSelector
--- PASS: TestDanglingServiceMonitor/TestNamespaceSelector
--- PASS: TestDanglingServiceMonitor/TestNamespacelessManifestStillMatches
--- PASS: TestDanglingServiceMonitor/TestNoDanglingServiceMonitors
--- PASS: TestDanglingServiceMonitor/TestOneDanglingServiceMonitorIsDangling
--- PASS: TestDanglingServiceMonitor/TestServiceMonitorEmpty
```

`go test ./pkg/templates/...` is green throughout.

### The reproduction from the issue, through the built binary

```
$ kube-linter lint repro.yaml --do-not-auto-add-defaults --include dangling-servicemonitor
(object: monitoring/application-metrics ...Kind=ServiceMonitor) no services found matching the
service monitor's label selector (app=application) and namespace selector ([])
Error: found 1 lint errors
```

and the control the issue names, the Service moved into `monitoring`:

```
No lint errors found!
```

### e2e

`tests/checks/dangling-servicemonitor.yml` gains that pair, so the fixture now produces five reports and the bats case asserts the fifth message and a count of five. Run against the built binary:

```
reports: 5
   ServiceMonitor: ... label selector (app.kubernetes.io/name=app) and namespace selector ([])
   ServiceMonitor: ... label selector (app.kubernetes.io/name=app) and namespace selector ([])
   ServiceMonitor: ... label selector () and namespace selector ([test2])
   ServiceMonitor: ... label selector (app.kubernetes.io/name=app1) and namespace selector ([test2])
   ServiceMonitor: ... label selector (app.kubernetes.io/name=monitoring-vs-application) and namespace selector ([])
```

AI-assisted (LLM used for drafting); every change verified and tested by me.


Eljees:
The one red check here is `codecov/project`, and it belongs to the repository rather than to this branch: codecov has had no report for `main` since 13 June 2024, so every pull request is compared against a base hundreds of commits behind. `codecov/patch` on this branch is green, and so are `build-and-test`, `kube-linter`, `test-sarif` and the Windows sanity test.

The measurement is in #1242, and #1263 carries the two parts of it that do not need repository settings access.

PR #1263, opened 2026-09-21
ci: make a rejected coverage upload visible and keep main's base fresh

Eljees:
The two parts of #1242 that do not need repository settings access.

### What is wrong today

Codecov has had no report for `main` since 13 June 2024, and `codecov/project` is consequently red on every open pull request that carries that context. From the last `push`-triggered run on `main` ([run 28674157419](https://github.com/stackrox/kube-linter/actions/runs/28674157419), `build-and-test`):

```
info  - Found 1 coverage files to report
info  - Upload queued for processing complete
error - Upload queued for processing failed: {"message":"Repository not found"}
```

The step still reported success, because `fail_ci_if_error` defaults to false. That is why two years passed unnoticed.

### The change

**`fail_ci_if_error` on both upload steps, except on `pull_request`.** A run that carries `CODECOV_TOKEN` now fails when its upload is rejected. Pull requests from forks have no access to the secret and upload tokenless, which is exactly the path that still works today, so they keep the tolerant behaviour. This is narrower than what I proposed in the issue, where I had suggested `true` everywhere; gating it this way avoids making every fork pull request depend on codecov's tokenless path staying healthy.

**A weekly `schedule` and `workflow_dispatch`.** `.github/workflows/build.yaml` last ran on a push to `main` on 3 July, although `main` has had commits since. They are dependabot auto-merges, performed by `.github/workflows/auto-merge.yaml` with `GITHUB_TOKEN`, and a push made with that token starts no further workflow run. So even with a working token, `main` stays dark for as long as its history is auto-merges. A weekly run gives the base something to be compared against, and `workflow_dispatch` makes it possible to refresh it on demand right after the token is rotated.

### What this does not do

Rotating `CODECOV_TOKEN` to the current upload token for `stackrox/kube-linter` needs repository settings access. Until that happens, this change makes the rejection visible on `main` rather than silent - which is the point, though it does mean the first scheduled run will be red while the token is still wrong. Happy to split that part out or hold this until the token is ready, whichever you prefer.

AI-assisted (LLM used for drafting); every change verified and tested by me.


Eljees:
`codecov/project` is red here too, and it stays red until the token is rotated: this branch changes when a rejected upload is reported, which does nothing for the stale base that the project check compares against. `codecov/patch` is green, as are `build-and-test`, `kube-linter`, `test-sarif` and the Windows sanity test.

One thing worth saying plainly: nothing in this pull request can be seen working until `CODECOV_TOKEN` is rotated. Merging it makes the next rejection visible on `main` instead of silent, and the scheduled run gives `main` a build it currently never gets.

Eljees:
Good catch on the ordering, and it was the more expensive of the two findings.

`fail_ci_if_error` on its own does exactly what the review says: the unit upload sits before the E2E and bats tests and before both binary artifacts, and `test-sarif` and `windows-sanity-test` both need `build-and-test`. So on every push to `main`, for as long as the token is wrong, a rejected upload would have cost the repository its whole validation rather than just telling someone the upload failed. That is a worse trade than the silence it was meant to replace.

Pushed `c9a27e5`: both uploads now carry an `id` and `continue-on-error: true`, so a rejection is recorded rather than fatal, and the last step in the job turns a recorded rejection into a job failure:

```yaml
      - name: Fail if a coverage upload was rejected
        if: >-
          github.event_name != 'pull_request' &&
          (steps.codecov-unit.outcome == 'failure' ||
           steps.codecov-bats.outcome == 'failure')
```

It runs after every test and every artifact upload, so the visibility the issue asked for is kept and nothing downstream is skipped to get it. Step order after the change, from the parsed workflow: unit tests, unit upload, E2E, bats, bats upload, Linux binary, Windows binary, and the check last.

The permissions finding from the first round is in `19c37ce`: `contents: read` on `build-and-test` and `windows-sanity-test`, plus `security-events: write` on `test-sarif` for the SARIF upload.

